PR #18, opened 2025-08-05
11 yaml loader and factory cycle detection reporting

gupichon:


JeanLucPons:
I propose the following patch which produces the error:
```
pyaml.exception.PyAMLException: /home/esrf/pons/pyaml/tests/config/sr/magnet_models/QF1AC01.yaml: Circular inclusion of sr/quadrupoles/QF1AC01.yaml
```
Still need to implemented access to line number mapping...

```python
     def __init__(self, filename:str, parent_path_stack:list):
        self.path:Path = get_root_folder() / filename
        self.files_stack = parent_path_stack if parent_path_stack is not None else []
        self.files_stack.append(self.path)

    # Expand condition
    def hasToExpand(self,value):
        expand = isinstance(value,str) and any(value.endswith(suffix) for suffix in accepted_suffixes)
        if expand:
            expandedPath:Path = get_root_folder() / value
            if any(expandedPath.samefile(p) for p in self.files_stack):
                raise PyAMLException(f"{self.path}: Circular inclusion: {value}")
        return expand

    # Recursively expand a dict
    def expand_dict(self,d:dict):
        for key, value in d.items():
            if self.hasToExpand(value):
                d[key] = load(value, self.files_stack)
            else:
                self.expand(value)

```

gupichon:
@JeanLucPons, have another look, I have updated tests and set a better exception message.

PR #35, opened 2025-10-16
Switch to name as identifier, removed typed magnet array

JeanLucPons:
Following discussion https://github.com/orgs/python-accelerator-middle-layer/discussions/26:
- PyAML objects are now identified using the name.
- Removed typed magnet array


gupichon (approved):
My only comment is that this should have been split into two separate merge requests — one for the IDs and one for the arrays.

Issue #6, opened 2025-07-28
Implement BPM device


gubaidulinvadim:
The idea is to do more or less the same as for magnets except that BPM names have to be unique in the lattice. An abstract `BpmModel` has to be implemented as for magnet.

However 2 cases appears:
Control system BPM having offsets (and tilt) implemented and BPM having only a position.
- First case 5 DevciceAccess are required in the pydantic model: h,v,offseth,offsetv,tiltangle (in that case, h,v reading from the CS already add the offsets and already rotate the coordinates)
- Second case only raw h and v come from the CS and offseth, offsetv and tilt angle have to be handled internally in pyaml.

We can focus first on the second case.

Concerning the CS, `RBpmArray`,`RWBpmTiltScalar` and `RWBpmOffsetArray` has to be implemented  [here](https://github.com/python-accelerator-middle-layer/pyaml/blob/main/pyaml/control/abstract_impl.py) and added [here](https://github.com/python-accelerator-middle-layer/pyaml/blob/main/pyaml/control/controlsystem.py)

Concerning simulators, `RBpmArray`, `RWBpmTiltScalar` and `RWBpmOffsetArray` has to be implemented  [here](https://github.com/python-accelerator-middle-layer/pyaml/blob/main/pyaml/lattice/abstract_impl.py) and added [here](https://github.com/python-accelerator-middle-layer/pyaml/blob/dc64a12702b85106e0b59d271514ed1c37fd573a/pyaml/lattice/simulator.py#L5)

`RBpmArray` for [h,v]
`RWBpmOffsetArray` for [offseth,offsetv]
`RWBpmTiltScalar` for the tilt angle

<img width="306" height="229" alt="Image" src="https://github.com/user-attachments/assets/26a9b802-f847-48dd-8e79-043918f36ebd" />

For arrays of BPM, it has also be implemented [here](https://github.com/python-accelerator-middle-layer/pyaml/tree/main/pyaml/arrays) following same  magnet arrays implementation.

gubaidulinvadim:
@kparasch, RW (read/write) implies that it is controllable.

PR #45, opened 2025-10-22
Feature implement bpm device

gubaidulinvadim:
First try at BPM implementation. Following Issue #6 and some discussions with Jean-Luc. I've only added the simplest tests for now.

### DIFF
--- a/pyaml/lattice/abstract_impl.py
+++ b/pyaml/lattice/abstract_impl.py
@@ -163,8 +163,8 @@ class RBpmArray(abstract.ReadFloatArray):
 
     # Gets the value
     def get(self) -> np.array:
-        index = self.lattice.get_refpts(self.element.FamName)
-        _, orbit = at.find_orbit(self.lattice, index)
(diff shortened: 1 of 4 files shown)


pyaml/lattice/abstract_impl.py
JeanLucPons: According to @gupichon modifications, an other field than `FamName` can be used for elements identification.
So you should pass it to the construct of your `RBpmArray` and use this field.
```python
index = self.lattice.get_refpts(self.element.FamName)
```

gubaidulinvadim:
I've made the corrections and also added a test with a nonzero orbit by setting one of the corrector magnet strength to 1e-6 in the test. FYI, I've created an issue in tango-pyaml https://github.com/python-accelerator-middle-layer/tango-pyaml/issues/21 but I'm not sure if it originates in tango-pyaml or in pyaml itself. If I use ` type: tango.pyaml.attribute_read_only` the configuration part fails, with `type: tango.pyaml.attribute` everything works.

PR #50, opened 2025-11-04
Improved exception handling

JeanLucPons:
This PR improves stack trace reading and exception handling.


tests/dummy_cs/tango-pyaml/tango/pyaml/controlsystem.py
```diff
@@ -17,6 +17,7 @@ class TangoControlSystem(ControlSystem):
     def __init__(self, cfg: ConfigModel):
         super().__init__()
         self._cfg = cfg
+        print(f"Creating dummy TangoControlSystem: {cfg.name}")
```
gubaidulinvadim: Maybe this should use logging instead of print but this is not too important now.

gubaidulinvadim (approved):

PR #51, opened 2025-11-04
20 feature implement tunemonitor

gubaidulinvadim:
Implements https://github.com/python-accelerator-middle-layer/pyaml/issues/20#issue-3402070077. The main purpose is to have a control system abstraction to measure the tune. Now we should be able to perform a tune response matrix measurement with a control system rather than a simulator. This will also allow the implementation and testing of use cases (that need tune measurement) on a real machine.


JeanLucPons (dismissed):
Thanks for the update.
Try to use relative import path.

Issue #46, opened 2025-10-22
Feature: Add hardware_name and model_name to elements


GamelinAl:
**Description, motivation and use case**
Both `hardware_name` and `model_name` need to be implemented. 
This will allow to make the link between the different names in pyAML:
- `name` = pyAML unique identifier.
- `hardware_name` =  control system name.
- `model_name` = lattice name.

**Proposed solution**
Add new properties to pyAML elements:
  - `hardware_name` corresponding to the reference string in the control system.
  - `model_name` corresponding to the element name in the lattice if any (`model_name` is not unique).

These properties are accessible in all control modes (live, design, ...).
The `hardware_name` is accessible if a control system has been defined in configuration.
The `model_name` is accessible if a simulator has been defined in configuration.

**Example**
```
Sextupole.name # returns 'SX.12.3'
Sextupole.hardware_name # returns 'AN01-AR/EM/SX.03'
Sextupole.model_name # returns 'SX.12'
```

GamelinAl:
If I understand correctly you mean that `hardware_name` can return either a str or a list of str. That would be fine with me, I don't think we should use `hardware_name` internally but it's mostly to inform the user.

The only thing we should take care is that, the element access in `ElementArray` by `hardware_name` search inside these list (https://github.com/python-accelerator-middle-layer/pyaml/issues/41).

PR #60, opened 2025-11-06
Dynamic elements arrays

JeanLucPons:
This PR propose dynamic arrays to answer [#46](https://github.com/python-accelerator-middle-layer/pyaml/issues/46#issuecomment-3495928705)

Here are few possible writing.
Improvement and ideas are welcome.
The selection code is located in pyaml/arrays/element_array.py

```python 
from pyaml.pyaml import pyaml,PyAML
from pyaml.configuration.factory import Factory
from pyaml.instrument import Instrument
from pyaml.arrays.element_array import ElementArray

ml:PyAML = pyaml("tests/config/EBSOrbit.yaml")
sr:Instrument = ml.get('sr')
ae = ElementArray("All",sr.design.get_all_elements())
acfm = ElementArray("AllCFM",sr.design.get_all_cfm_magnets(),use_aggregator=False)

bpmC5 = ae['BPM*'][10:20]
print(f"{type(bpmC5)}: {len(bpmC5)} items")  # All BPM C5
bpmc10 = ae['BPM*C10*'] # All BPM C10
print(f"{type(bpmc10)}: {len(bpmc10)} items")
magSHV = ae['SH*-V'] # All SH vertical corrector
print(f"{type(magSHV)}: {len(magSHV)} items")
magSH1A = ae['model_name:SH1A-*'] # All SH1A magnets (including the CFMs)
print(f"{type(magSH1A)}: {len(magSH1A)} items")
magSH1AC = acfm['model_name:SH1A-*'] # All SH1A magnets (CFMs only)
print(f"{type(magSH1AC)}: {len(magSH1AC)} items")
    
Factory.clear()
```
outputs

``` 
<class 'pyaml.arrays.bpm_array.BPMArray'>: 10 items
<class 'pyaml.arrays.bpm_array.BPMArray'>: 10 items
<class 'pyaml.arrays.magnet_array.MagnetArray'>: 96 items
<class 'pyaml.arrays.element_array.ElementArray'>: 129 items
<class 'pyaml.arrays.cfm_magnet_array.CombinedFunctionMagnetArray'>: 32 items
```

### DIFF
--- a/pyaml/arrays/element_array.py
+++ b/pyaml/arrays/element_array.py
@@ -9,7 +9,7 @@ import fnmatch
 
 class ElementArray(list[Element]):
     """
-    Class that implements access to a magnet array
+    Class that implements access to a element array
     """
 
     def __init__(self,arrayName:str,elements:list[Element],use_aggregator = True):
@@ -23,6 +23,8 @@ class ElementArray(list[Element]):
         elements: list[Element]
             Element list, all elements must be attached to the same instance of 
             either a Simulator or a ControlSystem.
+        use_aggregator : bool
+            Use aggregator to increase performance by using paralell access to underlying devices.
         """
         super().__init__(i for i in elements)
         self.__name = arrayName


gubaidulinvadim:
This looks mostly good to me. However, when I test I notice something strange:

```
In [15]: magSH1AC.strengths.get().shape
Out[15]: (97,)

In [16]: magSH1AC[0].strengths.get().shape
Out[16]: (4,)

In [17]: len(magSH1AC)
Out[17]: 32

In [18]: 4*32
Out[18]: 128
```
I do not understand what strengths of an array of combined function magnets are returned. For me, I would naively think that it will be a 2D array with a shape of (32, 4). A single ESRF combined function magnet returns array of (4,) corresponding to A0, B0, A1, B2. So, what is returned by a combined function array?

Also, there's a small issue for the user. The element array has no strength attribute because it can contain elements that are not magnets. But the user would want to have an array of, let's say, all H correctors. Some will be single function magnets and some will be combined function magnets. Then if I want to print the strength of all H correctors, I need to manually loop over all the magnets and have something like:

```
for m in magSH1A:
    if isinstance(m, CombinedFunctionMagnet):
        print(m.strengths[0].get())
    elif isinstance(m, SingleFunctionMagnet):
        print(m.strength.get())
```
Is this correct?

gubaidulinvadim:
I see, I didn't check if there are different types of combined function magnets. The strengths array in this example then is not very useful because it had to be flattened and there's no way to tell to which field component an array element belongs.

 For the user, it would definitely be good to have a way to change all H correctors (combined and not) in a single operation. For the naming and sorting-by-name part, this is already very good.

PR #67, opened 2025-11-13
3.10 compatibility

TeresiaOlsson:
Two new modules have been added where the import is not compatible with Python 3.10.


gubaidulinvadim (approved):
Sorry for that, both modules are from me.

PR #70, opened 2025-11-13
Fix missing root path for example.

TeresiaOlsson:
I have added the root path to the `esrf_tune_example_no_yaml.py` example since without it the csv files aren't found when running the example.

However, the example still doesn't work. I get the following error which I guess is because the devices try to connect to TANGO devices that don't exist.

```
PyAMLException: API_CorbaException: TRANSIENT CORBA system exception: TRANSIENT_NoUsableProfile Origin: void Tango::Connection::connect(const string&) at (/src/cppTango/src/client/devapi_base.cpp:635) Severity: ERR
```

### DIFF
--- a/examples/ESRF_tune_example/esrf_tune_example_no_yaml.py
+++ b/examples/ESRF_tune_example/esrf_tune_example_no_yaml.py
@@ -7,10 +7,23 @@ from pyaml.arrays.magnet import Magnet,ConfigModel as MagnetArrayConfigModel
 from tango.pyaml.controlsystem import TangoControlSystem,ConfigModel as ControlSystemConfig
 from tango.pyaml.attribute import Attribute,ConfigModel as AttributeConfig
 from tango.pyaml.attribute_read_only import AttributeReadOnly,ConfigModel as AttributeReadOnlyConfig
+from pyaml.configuration import set_root_folder
+import os
 
 
 import numpy as np
 import time
+
+# Get the directory of the current script
+script_dir = os.path.dirname(__file__)
+
+# Go up one level and then into 'data'
+relative_path = os.path.join(script_dir, '..', '..', 'tests')
+
+# Normalize the path (resolves '..')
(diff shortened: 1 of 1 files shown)


TeresiaOlsson:
Okay. I thought the purpose of the example was to show that what I want to be able to do can already be done with the current implementation.

JeanLucPons:
For me your patch is ok:
```
[ubuntu20acu.pons] > python esrf_tune_example_no_yaml.py
Tango control system binding for PyAML was already initialized with name 'live' and TANGO_HOST=ebs-simu-3:10000
Tune=Value(0.2921267341037133, quality='VALID', timestamp='2025-11-14 07:07:14.607817'), Value(0.40879552981619777, quality='VALID', timestamp='2025-11-14 07:07:14.608263')
Tune=Value(0.38952758430038753, quality='VALID', timestamp='2025-11-14 07:07:17.672885'), Value(0.3614632411496573, quality='VALID', timestamp='2025-11-14 07:07:17.673515')
```

To use without control system:

```
[ubuntu20acu.pons] > pwd
/home/esrf/pons/pyaml
[ubuntu20acu.pons] > pip install tests/dummy_cs/tango-pyaml
[ubuntu20acu.pons] > cd examples/ESRF_tune_example/
[ubuntu20acu.pons] > python esrf_tune_example_no_yaml.py
Creating dummy TangoControlSystem: live
Tune=Value(0.0, quality='VALID', timestamp='2025-11-14 07:10:21.813228'), Value(0.0, quality='VALID', timestamp='2025-11-14 07:10:21.813282')
MultiAttribute.get(4 values)
MultiAttribute.get(4 values)
MultiAttribute.set(4 values)
Tune=Value(0.0, quality='VALID', timestamp='2025-11-14 07:10:24.813651'), Value(0.0, quality='VALID', timestamp='2025-11-14 07:10:24.813730')
```

Issue #28, opened 2025-10-06
Add an API to instantiate an instrument without configuration file


gupichon:
**Description:**
Currently, an instrument can only be created from a YAML configuration file.
It should also be possible to instantiate an instrument programmatically using a direct API call.

**Example use case:**
Developers may want to:
* Build instruments dynamically (e.g., from in-memory definitions or database records)
* Use PyAML in scripts or tests without maintaining a YAML file on disk

**Expected feature:**
Provide a Python API such as:
```python
from pyaml import Instrument

instr = Instrument.from_dict({
    "name": "ring",
    "devices": [...],
})
```

TeresiaOlsson:
Why does it need to include a dict? I thought the user could just create objects for the required configuration dataclass/es and then use them to generate the objects with logic since the config classes already exist?

gupichon:
Sounds good — we can support both approaches. I initially leaned toward a plain dict because it mirrors the YAML schema and is straightforward to implement. That said, building from typed objects (or a configuration object) is also feasible. We can keep the dict-based API for parity with YAML and add an object-based builder, both sharing the same validation layer.

JeanLucPons:
It should be possible to write something like:
```python
control = TangoControlSystem(name="live",tango_host="ebs-simu-3:10000")
qfCurve = CSVCurve(file="config/sr/magnet_models/QF1_strength.csv")
qAtt = Attribute(attribute="srmag/vps-qf1/c01-a/current")
qModel = LinearMagnetModel(name="QF1",curve=qfCurve,calibration_factor=1.00504,powerconverter=qAtt,unit="1/m")
quad = Quadrupole(model=qModel,peer=control)
quad.strength.set(0.8)
```

TeresiaOlsson:
It should also include data validation in the same way as you get when using the yaml file.

PR #71, opened 2025-11-13
Refactor config layer

TeresiaOlsson:
I'm attempting to refactor the configuration layer based on my suggestion in the `aml` repository and the comments from the maintainers meeting.

In the first step:

- I have removed the `PyAML` class and moved everything related to loading from a file into the fileloader module.
- I removed the dictionary of instrument since it's not required since it's easier to just load each instrument separately. Especially if people want to initialize all devices when loading and be able to keep the configuration for different instruments at different root paths.
-  I tried to add a similar interface as in pyAT where there are two options for creating a lattice, `load_lattice` or create a `Lattice` object directly. In this suggestion there is now `load_instrument` to load from file or you can create an `Instrument` object directly.

This aim to solve issue #28.

### DIFF
--- a/tests/test_factory.py
+++ b/tests/test_factory.py
@@ -49,4 +49,4 @@ def test_error_cycles(test_file):
 
     assert "Circular file inclusion of " in str(exc.value)
     if not test_file.endswith(".json"):
-        assert "at line 23" in str(exc.value)
+        assert "at line 21" in str(exc.value)


gubaidulinvadim:
For your circular import, it is caused (from what I understand) by wanting to specify the type (Instrument). In some packages, like numpy, there's a special typing module to avoid this.

JeanLucPons:
For solving this kind of circular import you can:
But you still need to use type annotation as str.

```python
from typing import Union, TYPE_CHECKING

if TYPE_CHECKING:
    from pyaml.instrument import Instrument

def load_instrument(filename:str, paths_stack:list=None) -> "Instrument":
```

simoneliuzzo:
Also for me the single instrument (or Accelerator, ok for me as well) per yaml config file is also ok. I can help if needed, let me know @TeresiaOlsson

PR #72, opened 2025-11-14
Updated build of tango dummy-cs

JeanLucPons:
This PR solves issue when switching between dummy-cs and true tango cs.


TeresiaOlsson:
> I actually fail to reproduce the issue either here or on the main branch. @TeresiaOlsson, could you check that there's no conflict anymore with tango?

I don't understand. What should I check? I have had no issues with dummy-cs because I never tried to run any of the tests.

simoneliuzzo (approved):

PR #74, opened 2025-11-14
first minimal ophyd-async implementation

### LINKED ISSUE #21: EPICS bindings / ophyd-async
# Description
A very basic binding that just implements DeviceAccess (read/write a float value), DeviceAccessList (read/write a list of float value) and ControlSystem (control system initialization) would be already very nice.
Tango or EPICS bindings should just implement the abstract classes present in PyAML.
Then to install pyaml you can install either pyaml-tango or pyaml-epics which will install all. If you install only pyaml alone, you will be limited to model/design access only.
For Tango as the set/get and readback are accessible by just one attribute, the config will be different for Epics and Tango.

# Config example

```
# Here tango.pyaml.attribute inherits from pyaml.control.deviceaccess.DeviceAccess
type: tango.pyaml.attribute 
attribute: srmag/vps-qf1/c01-a/current
unit: A

# A config for epics could be (for instance):
type: epics.pyaml.pv
set_pv: SR:MAGNET:PS:SET_CURRENT
readback_pv: SR:MAGNET:PS:CURRENT
unit: A

# Tango control system initialization
controls:
    tango_host: ebs-simu-3:10000
    name: live
```

For the DeviceAccessList, there is no particular config. It is just a list of DeviceAccess that is used internally for arrays. For Tango, we use asynchronous call to read values in parallel.

yhidaka:
@JeanLucPons @gubaidulinvadim @gupichon @TeresiaOlsson Please take a look at my first minimal-effort implementation that allows both EPICS and Tango access based on `ophyd-async` signals, and let me know your opinion. See [README](https://github.com/python-accelerator-middle-layer/pyaml/blob/yh_integrate_ophyd_async/examples/ESRF_tune_example/ophyd_async/README.md) for the instructions.

I don't have a digital twin to make tune changes occur, so I just checked the get/set methods work in both the Jupyter notebook and the Python script, with some random values.

Also, I don't have much experience with `asyncio`. So, most likely professional software engineers need to re-implement async handling in a more proper manner. My focus was just to make it work in Jupyter as well as the integrated terminal and the debugger for vscode.


JeanLucPons:
Thanks yoshi.
The only things I dislike is that now pyaml is control system dependent as you have both Tango and Epics related code inside.
I would be nice to move `pyaml/control/signal` outside.
However nice work !
I also plan to refurbish the control system interface which prevents for the moment to connect to several control system instance.

Now if we decide to put control system dependent code inside pyaml then we can also move the actual tango backend inside pyaml. It relies on fast c++ asynch io that we need because at ESRF thousands of setpoints are launch at each orbit correction steps.

JeanLucPons:
I finally managed to make it work using this ugly patch !
Inpisred from [here](https://blueskyproject.io/ophyd-async/main/_modules/ophyd_async/testing/_wait_for_pending.html)

```python
    async def async_set(self, value):
        await asyncio.sleep(0) # Yield, may need several
        return await _recover_once(lambda: self._run_set(value),
                                   self._w_sig.connect, getattr(self._w_sig, "__rebuild__", None))

    def set(self, value):
        """Synchronous wrapper around `async_set()`."""
        r = arun(asyncio.gather(self.async_set(value))) # Use gather (without it fails !)
        print(f"Set {self._w_sig}={value} {r}")
        return r
```

However returned status is *pending* !!?
```
Set <ophyd_async.core._signal.SignalW object at 0x7f68b8448530>=93.67912489221612 [<AsyncStatus, task: <coroutine object SignalW.set at 0x7f68b33b3010>, pending>]
Set <ophyd_async.core._signal.SignalW object at 0x7f68b33c0d70>=94.79924472193656 [<AsyncStatus, task: <coroutine object SignalW.set at 0x7f68b33b3890>, pending>]
Set <ophyd_async.core._signal.SignalW object at 0x7f68b33c11f0>=109.2133545845729 [<AsyncStatus, task: <coroutine object SignalW.set at 0x7f68b33b39a0>, pending>]
Set <ophyd_async.core._signal.SignalW object at 0x7f68b33c1610>=106.39116921366295 [<AsyncStatus, task: <coroutine object SignalW.set at 0x7f68b33b3120>, pending>]
```

JeanLucPons:
Here is typical behavior (not always reproductible but frequent) I can observe without the patch.
The last quad (QD2E-C01) is not set when `run_unit_complete()` ends. Then the task is then blocked by the `time.sleep()` and finally the quad is applied when the tune is read (when the program enters again in the event loop).

image

Here i used a waiting time of 10sec.

```python
strs = quadForTuneLive.strengths.get()
strs += np.matmul(correctionmat, [0.1, 0.05])  # Ask for correction [dqx,dqy]
quadForTuneLive.strengths.set(strs)

time.sleep(10)

print(f"Tune-X={qxAtt.readback()}, {qxAtt.RB.get()}")
print(f"Tune-Y={qyAtt.readback()}, {qyAtt.RB.get()}")
```

gubaidulinvadim:
This is very good! I would agree with Jean-Luc that we should move it to a separate package (something like "ophyd-pyaml"). This is also important because we need a way to implement DOOCS at some point. This can be done in two ways: one is to implement ophyd-doocs, and the other is to do a similar thing to tango-pyaml. We need to keep this flexibility for now.

gubaidulinvadim:
@yhidaka Ah, I didn't notice that the tango and epics imports only occur when you create some mock-devices/pvs. Then it's a question of having or not having an ophyd-async dependency.

JeanLucPons:
> As for moving the interface module outside of `pyaml`, I'm not sure that is a good idea. `ophyd-async` already handles the control system abstraction part. I implemented the interface such that tango facilities do not have to install epics dependencies (and vice versa) even with the interface embedded in `pyaml`. So, to me, it seems redundant and overkill to separate it out as an external package.

I'm not happy with the ophyd-async implementation that relies on Tango asyncio. It is too slow. I want to be able to use Tango asynchronous calls based on C++. I don't want to have to maintain code inside pyaml that I cannot fully test. Handling event loop enter/exit is tricky and this is not the good way to use an async API.

kparasch:
I think having `ophyd-async` as an external package is not a bad idea. For DOOCS I know I can develop a `doocs-pyaml` package and while even though it is likely possible, I don't know how easy it will be to contribute to `ophyd-async` which is in another organisation with their own governance. 

Apart from DOOCS, having ophyd-async as an external package protects us from major code refactoring in case `ophyd-async` ever stops to be maintained or changes its API.

Finally, if another lab with yet another control system would find interest in using `pyaml` in the future, I think it would discourage them if we told them that first they need to write `ophyd-async`-compatible bindings instead of their own.

yhidaka:
@JeanLucPons  As I cannot reproduce the racing condition on my end, could you please try changing the `set` method as follows with the additional `_complete_set` method in `container.py`? I think this will ensure the coroutine gets executed before `time.sleep` blocks this thread.

```python
class Setpoint:
    ...
    async def _complete_set(self, value):
        status = await self.async_set(value)
        await status  # Wait for completion before returning
        return status

    def set(self, value):
        """Synchronous wrapper around `async_set()`."""

        status = arun(self._complete_set(value))
        return status
```

Obviously, if you have multiple setpoints, we should collect their coroutines first and gather on them. But, doing this one by one synchronously also needs to work.

gubaidulinvadim:
> All we need to agree is to use common APIs, which I think we should follow those of ophyd-async, as they must have thought through.

@yhidaka, could you expand on this? I do not understand the argument of "they must have thought through". Then the same can apply to saying that we should follow tango devices and tango attributes as common APIs. So, to me, this alone is a weak argument. Especially, if one is not familiar with ophyd-async.

JeanLucPons:
> @JeanLucPons , I'm sorry, I don't think the "fully testable" part is clear still.

You need to have something that is "independent" of underlying layers with well defined standards.
For instance matlab has nothing to do with Tango or Epics but nothing prevent to write extensions to use Tango in matlab.
pyaml must follow the same model.
The work is rather to define the interface for extensions rather than integrating all possible underlying layers inside the code.

yhidaka:
> @yhidaka, could you expand on this? I do not understand the argument of "they must have thought through". Then the same can apply to saying that we should follow tango devices and tango attributes as common APIs. So, to me, this alone is a weak argument. Especially, if one is not familiar with ophyd-async.

It's not the same. The APIs for tango were most likely created only for Tango without regard to any compatibility with EPICS (If this is not the case, please point me to a relevant source.) However, `ophyd-async` APIs were, from the beginning, designed to be compatible with multiple control systems (specifically EPICS and Tangos as a start) by utilizing a backend-based design. Signal objects in ophyd-async do not contain control system-specific logic.

This does not prevent us from reinvestigate how to provide a common interface. But, I think everybody involved in this aspect of pyAML at least should try to understand what their wisdoms and deficiencies are.

JeanLucPons:
>  However, `ophyd-async` APIs were, from the beginning, designed to be compatible with multiple control systems (specifically EPICS and Tangos as a start) by utilizing a backend-based design. Signal objects in ophyd-async do not contain control system-specific logic.

It is great to define standard but not to add implementation inside. Otherwise this in one layer more. It is not always easy to have support from Tango itself so if we add one more layer that contains code, pyaml will be more difficult to maintain. There is already lots of Tango related code in ophyd-async. It would be great to provide a clear documentation on how to implement an Ophyd (asynch or not) implementation from scratch.

JeanLucPons:
> @JeanLucPons As I cannot reproduce the racing condition on my end, could you please try changing the `set` method as follows with the additional `_complete_set` method in `container.py`? I think this will ensure the coroutine gets executed before `time.sleep` blocks this thread.

Nice, it works and the patch is clean !
I did few tests without problem... before only ~ 1 out of 4 tests was ok.

image

PR #77, opened 2025-11-18
Added use fast yaml loader option

JeanLucPons:
This PR adds an option to use fast yaml loader:
```python
#        use_fast_loader : bool
#            Use fast yaml loader. When specified, no line number are reported in case of error, 
#           only the element name that triggered the error will be reported in the exception)
sr = Accelerator.load("tests/config/EBSOrbit.yaml",use_fast_loader=True)
```


gubaidulinvadim (approved):
For me, this is good. It passes all the tests. The only thing I do not understand is if we have a fast c-loader, what is the point of having a slow loader (and the slow one being the default option)?

PR #78, opened 2025-11-18
2 handle serialized magnets

### LINKED ISSUE #2: Handle serialized magnets
Handle serialized magnets:
A pyaml element can have several associated AT elements.

https://github.com/python-accelerator-middle-layer/pyaml/blob/0e306fbf0f058d0159dc976d148b5d23a183d296/pyaml/lattice/abstract_impl.py#L7

gupichon:
## Description

Handling serialized magnets. Changing one of the serie changes all of them, by passing by the model as the current is common. Specific corrections factors can be set by element.

## Related Issue
- Fixes #2 

- [ ] Serialized magnets model
- [ ] Linear model for serialized magnet with he possibility to fine tune each element with a specific curve and/or factors.

## Testing
 - [ ] test_serialized_magnets.py


JeanLucPons:
> - magnets sharing the same name: 1ps, 1 magnet model, 1 pyaml magnet for the family 


For this use case (classic serialized magnet or split magnet) you have just to handle correctly the loop over all elements instead of only the first one in both `RWHardwareScalar` and `RWStrengthScalar` class:

```python
    # Gets the value
    def get(self) -> float:
        return self.__poly[self.__polyIdx] * self.__elements[0].Length

    # Sets the value
    def set(self, value:float):
        self.__poly[self.__polyIdx] = value / self.__elements[0].Length
```


> - magnets having different models and nanes: 1ps, n magnet models, 1 pyaml magnet for the family and n pyaml (virtual) magnet -> RWHardware , RWStrength (if you want)
 
For this use case, your serialized_magnet class is close to the need.

simoneliuzzo:
Dear @gupichon and @JeanLucPons ,

I agree with the comments above concerning: 
- magnets sharing the same name: 1ps, 1 magnet model, 1 pyaml magnet for the family

For 
- magnets having different models and names: 1ps, n magnet models, 1 pyaml magnet for the family and n pyaml (virtual) magnet -> RWHardware , RWStrength
I find this possibility very useful (This is actually what we will have for our booster soon).
Could it be extended to the case of different number of turn in the coils and different lengths magnets on the same PS?

gupichon:
It's corrected

gupichon:
It should be ok now.

JeanLucPons:
It seems fine.
thanks.

JeanLucPons (approved):

Issue #80, opened 2025-11-19
BPMArray is not linked correctly with simulator


kparasch:
I tried to read the orbit from the `BPMArray` (called 'BPM') using the EBSOrbit.yaml configuration. It works using `sr.live` but not when using `sr.design` getting the error below. I haven't tried yet to look more what is the exact issue, but maybe someone more familiar knows already what is the issue or if I am doing something wrong.

This is a minimal script to reproduce it:
```
from pyaml.accelerator import Accelerator

config_path('pyaml/tests/config/EBSOrbit.yaml') # replace accordingly
sr = Accelerator.load(config_path)

sr.live.get_bpms('BPM').h.get() # ok
sr.design.get_bpms('BPM').h.get() #throws error
```

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
File ~/workspace/010_pyaml/pyaml/examples/ESRF_ORM_example/esrf_orm_example.py:12
      9 sr = Accelerator.load(config_path)
     11 sr.live.get_bpms('BPM').h.get() # ok
---> 12 sr.design.get_bpms('BPM').h.get()
     14  # class ConfigModel(BaseModel):
     15  # 
     16  #     model_config = ConfigDict(arbitrary_types_allowed=True,extra="forbid")
   (...)
     35  # 
     36  # interface = pySCInterface(ConfigModel(element_holder=sr.design))

File ~/workspace/010_pyaml/pyaml/pyaml/arrays/bpm_array.py:44, in RWBPMSinglePosition.get(self)
     42     return np.array([b.positions.get()[self.__idx] for b in self.__bpms])
     43 else:
---> 44     return self.__aggregator.get()

File ~/workspace/010_pyaml/pyaml/pyaml/lattice/abstract_impl.py:202, in BPMHScalarAggregator.get(self)
    201 def get(self) -> np.array:
--> 202     _, orbit = at.find_orbit(self.__lattice, refpts=self.__refpts)
    203     return orbit[:, 2]

AttributeError: 'BPMHScalarAggregator' object has no attribute '_BPMHScalarAggregator__lattice'
```

PR #81, opened 2025-11-19
Access the correct variables in BPMHScalarAggregator and BPMVScalarAggregators

kparasch:
Because they are inheriting a private variable, their names are mangled. It is ugly to access the variables like this. I would like advice how you think we could do this better.

Changes:
 - Add test that originally fails when accessing bpm aggregators' `h` and `v` properties.
 - Move `__lattice` and `__refpts` of `BPMScalarAggregator` to public namespace.
 - Reverse `BPMHScalarAggregator`, `BPMVScalarAggregator` as they were doing the opposite.

Solves #80 .

### DIFF
--- a/pyaml/lattice/abstract_impl.py
+++ b/pyaml/lattice/abstract_impl.py
@@ -158,11 +158,11 @@ class BPMScalarAggregator(ScalarAggregator):
     """
 
     def __init__(self, ring:at.Lattice):
-        self.__lattice = ring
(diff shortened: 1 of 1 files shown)


JeanLucPons:
I agree that it is not very user friendly and that it is not supposed to happen. I would suggest to either use a `get_lattice()` function or to move the field to public in the super class.

PR #82, opened 2025-11-20
Orbit correction and Orbit Response Matrix measurement through pySC.

kparasch:


JeanLucPons:
> @JeanLucPons I don't have the separation between h and v yet in pySC. A quicker fix to see if this is indeed the issue could be if you try:
> 
> ```
> response_matrix.matrix[:,len(response_matrix.input_names)//2:] *= -1
> ```
> 
> or if you pass an empty vcorr array to calculation of the ideal ORM, which you can then use with a gain=-1

It is worse. Now the 2 planes are failling....
pyaml is not using KickAngle whic is [definitely wrong](https://github.com/atcollab/at/pull/1001#issuecomment-3435604006) ;)
I can try to recompute the ORM and trick set strength.

JeanLucPons:
Yes yes, i agree. I will add a sign in `PolynomInfo` as a quick fix.

kparasch:
@gubaidulinvadim Tests should be ok now

kparasch:
The control access is very minimal actually in this PR, so if it is not compatible with Jean-Luc's refurbishment, it will not be difficult to adapt.

pySC does not wait for anything :) The only place where the set is actually done is on the `pySC_interface.py` file. Now it's using the `set` of pyAML but indeed we should go for 'set_and_wait' when it is implemented. Personally I don't see a good way to avoid putting waiting times. I think we should review this topic more at some point. Even after the 'set_and_wait' function has finished (based on power supply readback value I assume) we will have to at least handle decay of eddy current fields and other transients, as well as polling rate and integration time of electronics for BPMs.

In the ESRF tests, we put some waiting times as attributes in the equivalent pySC_interface which we would tweak.

JeanLucPons (approved):
Nice work.
Thanks again Kostas !

PR #84, opened 2025-11-24
Fix sign issue for horizontal kick angle

JeanLucPons:
This PR solves sign issue for horizontal kick angle convention.
See [#82](https://github.com/python-accelerator-middle-layer/pyaml/pull/82)


gubaidulinvadim:
It looks good, but the sign convention should be specified in the configuration file, not in the source code.

JeanLucPons:
Yes i would be in favor on fixing one.

JeanLucPons:
But i also think that it is OK for the moment.
Discussion around convention will take ages !

gubaidulinvadim:
@JeanLucPons , did you fix your issue? When I run locally the pre-commit I had no issue just right now. It only fixed the trailing white spaces for me. And it was only in docs/, tests/ and .github/ folders. Those are maybe not included in the actual pre-commit hook and CI/CD. I'm not sure.

Edit: failed ruff CI/CD pipeline from your commits above only has vcorrector.py and hcorrector.py.

kparasch (approved):

PR #92, opened 2025-12-10
Attachment of orbit correction to simulator/control system and generic ResponseMatrix pyAML object

kparasch:


kparasch:
@JeanLucPons 
The assumption on the ordering will "usually" work depending how the response matrix was built, because the keys in trims comes from `response_matrix.input_names`.

A more robust way would be to iterate over names of the array:

```
# join h and v
trims = {}
for key in trims_h.keys():
    trims[key] = trims_h[key]
for key in trims_h.keys():
    trims[key] = trims_v[key]


arr_names = self._hvcorr.names()
data_to_send_hv = self._hvcorr.strengths.get()
for idx,name in enumerate(arr_names):
     data_to_send_hv[idx] += trims[name] * gain
```

and we should probably also make a check that all correctors from trims will be set in the arrays. 

```
for name in trims.keys():
    assert name in arr_names

self._hvcorr.strengths.set(data_to_send_hv)
```

Do you want to merge the branches and one of us continues?

JeanLucPons:
I rebased and merged my mods.
I let you optimize the join of h and v.
The sample below is working:

```python
sr = Accelerator.load("tests/config/EBSOrbit.yaml",use_fast_loader=False)
sr.live.orbit.correct()
```

kparasch:
@JeanLucPons  This is basically finished except one thing:

```
    def post_init(self):
        self._hcorr = self._peer.get_magnets(self._cfg.hcorr_array_name)
        self._vcorr = self._peer.get_magnets(self._cfg.vcorr_array_name)
        hvElts = []
        hvElts.extend(self._hcorr)
        hvElts.extend(self._vcorr)
        self._hvcorr = MagnetArray("HVCorr", hvElts)
```

Here we create a magnet family with a hardcoded name "HVCorr". I think we should not do this so that it doesn't break if someone decides to define a magnet family called "HVCorr". 
I don't have any good solution.

PR #95, opened 2025-12-11
Added facility and updated name to machine.

TeresiaOlsson:
I have done three changes:

- Added a field `facility` to the `Accelerator` that could be used in metadata to know at which facility a specific dataset was produced at. It's also needed for labs that run more than one facility.

- Suggested to change `name` in `Accelerator` to `machine` to make it easier to understand what the field means when there is also the `facility` field.

- Cleaned up remains from when `Instrument` was changed to `Accelerator`.

### DIFF
--- a/pyaml/accelerator.py
+++ b/pyaml/accelerator.py
@@ -1,5 +1,5 @@
 """
-Instrument class
+Accelerator class
 """
 
 from pydantic import BaseModel, ConfigDict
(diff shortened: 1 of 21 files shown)


gubaidulinvadim:
@TeresiaOlsson , for me the facility should not be mandatory. I don't like "machine" instead of "name". You have an accelerator object with a property "name", for me it's clear it refers to accelerator name. But this is just a personal preference.

PR #96, opened 2025-12-11
Update installation instructions in README.md

TeresiaOlsson:
We need better instructions in the readme for how to install the package depending on what you want to do.

I started to make a separation of user installation and development installation but I don't think the instructions are really good yet. Missing information:

- Do you also need to update the submodule for a user installation or this is just needed to be able to run the tests if you are a developer?

- How do you install the optional bindings for tango, ophyd-async etc. The current instructions where you need to clone a separate repository manually seems to complicated for an ordinary user who want to test to provide feedback. I think we need to get these optional dependencies into the pyproject.toml. But I don't know what the correct procedure is to install those dependencies. Can someone who know perhaps add that?

- Also the developer instructions are a bit unclear. What does "For tests, you may want to install dummy-cs/tango available in tests/dummy-cs/tango"? Do you need to install it or not to be able to run the tests? I thought you had to?

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,27 +5,14 @@
 
 Disclaimer: the pyAML software is still under development.
 
-This repository is pyAML core. It is control system independent and provides the main functionality of pyAML (device abstraction for magnets, bpms, etc.; access to simulator (pyAT) and abstract implementation for control system). It is intended to be used together with a package that implements control system specific interface (for example, tango-pyaml or pyaml-cs-oa). If you install tango-pyaml or pyaml-cs-oa it will install pyaml automatically (it is listed as a mandatory dependency for this packages).
+This repository is pyAML core. It is control system independent and provides the main functionality of pyAML (device abstraction for magnets, bpms, etc.; access to simulator (pyAT) and abstract implementation for control system). It is intended to be used together with a package that implements control system specific interface.
 
+If you want to only tests for your machine and not develop, you shouldn't install this package. Instead go to the package for the bindings and install that one. It will install this package automatically.
 
-#### User Installation
+Available packages for bindings:
 
-1. Clone the repository (optional)
-2. Create a virtual environment and activate it
-3. Install the package via pip
-
-   ```
-   cd pyaml
-   pip install .
-   # pip install accelerator-middle-layer #if you want to install from PyPi.org
-   ```
-   
-4. If you want to try the examples using the TANGO bindings you also need [tango-pyaml](https://github.com/python-accelerator-middle-layer/tango-pyaml).
-   Clone that repository and install the package inside the same virtual environment as the `pyaml` package.
-   tango-pyaml will automatically install pyaml with a correct version, so step 3 can be skipped.
-   ```
-   pip install .
-   ```
+TANGO: [tango-pyaml](https://github.com/python-accelerator-middle-layer/tango-pyaml)  
+TANGO or EPICS: [pyaml-cs-oa](https://github.com/python-accelerator-middle-layer/pyaml-cs-oa)
 
 #### Developer Installation
 
@@ -47,15 +34,13 @@ This repository is pyAML core. It is control system independent and provides the
    pip install -e .[dev]
    pre-commit install
    ```
-   
-5. If you want to try the examples using the TANGO bindings you also need [tango-pyaml](https://github.com/python-accelerator-middle-layer/tango-pyaml).
-   Clone that repository and install the package inside the same virtual environment as the `pyaml` package.
-   tango-pyaml will automatically install pyaml, so step 3 can be skipped.
-   
-7. If you want to run tests manually, you may want to install dummy-cs/tango available in
-   tests/dummy-cs/tango
 
-   dummy-cs package is the simplest emulation that allows to check the interface to the control system. (The implemented control system doesn't do anything, it is only intended for tests during the development)
+5. If you want to try the examples using the control system bindings you also need to install those. See:
+
+   TANGO: [tango-pyaml](https://github.com/python-accelerator-middle-layer/tango-pyaml)  
+   TANGO or EPICS: [pyaml-cs-oa](https://github.com/python-accelerator-middle-layer/pyaml-cs-oa)
+ 
+6. If you want to run tests manually using the TANGO bindings without requiring a live machine you can also install the the Dummy TANGO control system available in tests/dummy-cs/tango. It is a simple emulation that allows to check the interface to the control system. The implemented control system doesn't do anything but it is only intended for tests during the development.
 
 
 #### Documentation


TeresiaOlsson:
Sorry, I still don't understand. For the instructions I wanted to know if people need to install the dummy backend to run the pytests. I thought that was the case?

gubaidulinvadim:
@TeresiaOlsson I've updated the readme a little bit. The pyaml repository is control system independent. And in the current implementation, I think, we still cannot create an Accelerator object without any control system attached (@JeanLucPons please correct me if I'm wrong here). So, the end user (when development is finished) will only need to install either tango-pyaml; pyaml-cs-oa where pyaml is listed as a mandatory dependency. 

dummy-cs package could be installed optionally (for developers) to work without any connection to a real control system. You can run tests without it, pytest is using fixtures and will install and uninstall it for you (if you run pytest). It is only needed to be installed if you want to have an interactive session in ipython/jupyter notebook to play around with the interface.

I cannot answer any pySC questions @kparasch, I didn't have the time yet to play with it. For tests, I think we should follow the same approach as with dummy-cs and make it install/uninstall automatically by pytest.

@TeresiaOlsson , yes I think we could include dummy-cs in pyproject.toml under development dependencies but this is not mandatory. (And I'm not sure if pyproject.toml can install local packages without going though pypi)

kparasch:
It is true that in the current implementation pySC is a mandatory dependency. I can make it optional, though I would find it misleading at this stage because there is no alternative solution and someone who wants to just correct the orbit will need to go look up for a special paragraph of the user guide.

Some points that I think are important to make:
- The interfacing to pySC is a shortcut with little code and to me pySC does not need to be a permanent solution.
- We can at any time decide that we don't want to use pySC and we can use something else, re-write pySC from scratch or even copy the relevant source code of pySC into pyAML.
- Unlike having epics, tango, doocs, ophyd-async or bluesky as mandatory dependencies, pySC does not affect the design of the rest of pyAML.

The biggest advantage I see now is that it allows us to develop how we want the configuration of the orbit correction to look like and how the user can interact with it, without locking into a specific technology.

TeresiaOlsson:
> @TeresiaOlsson I've updated the readme a little bit. The pyaml repository is control system independent. And in the current implementation, I think, we still cannot create an Accelerator object without any control system attached (@JeanLucPons please correct me if I'm wrong here). So, the end user (when development is finished) will only need to install either tango-pyaml; pyaml-cs-oa where pyaml is listed as a mandatory dependency.
> 
> dummy-cs package could be installed optionally (for developers) to work without any connection to a real control system. You can run tests without it, pytest is using fixtures and will install and uninstall it for you (if you run pytest). It is only needed to be installed if you want to have an interactive session in ipython/jupyter notebook to play around with the interface.
> 
> I cannot answer any pySC questions @kparasch, I didn't have the time yet to play with it. For tests, I think we should follow the same approach as with dummy-cs and make it install/uninstall automatically by pytest.
> 
> @TeresiaOlsson , yes I think we could include dummy-cs in pyproject.toml under development dependencies but this is not mandatory. (And I'm not sure if pyproject.toml can install local packages without going though pypi). The easiest for the end user is also to have everything on pypi.org. Then you can just do pip install package_name. And the end user would only need to do either pip install tango-pyaml or pip install pyaml-cs-oa.

Very good. I just wanted some good instructions so the people who are not in the maintainers group don't get confused about what they need to install to test and provide feedback from their labs. For developers I think it's perfectly fine that there are mandatory requirements and that you might need to install control system dependent packages, pySC etc. 

I think there is an option in a pyproject.toml to give a dependency of a github repository with the purpose to use doing development when you need a package that hasn't been published yet but I haven't had time to test that myself so I don't know how well it works. I can try it.

TeresiaOlsson:
> It is true that in the current implementation pySC is a mandatory dependency. I can make it optional, though I would find it misleading at this stage because there is no alternative solution and someone who wants to just correct the orbit will need to go look up for a special paragraph of the user guide.
> 
> Some points that I think are important to make:
> 
>     * The interfacing to pySC is a shortcut with little code and to me pySC does not need to be a permanent solution.
> 
>     * We can at any time decide that we don't want to use pySC and we can use something else, re-write pySC from scratch or even copy the relevant source code of pySC into pyAML.
> 
>     * Unlike having epics, tango, doocs, ophyd-async or bluesky as mandatory dependencies, pySC does not affect the design of the rest of pyAML.
> 
> 
> The biggest advantage I see now is that it allows us to develop how we want the configuration of the orbit correction to look like and how the user can interact with it, without locking into a specific technology.



> It is true that in the current implementation pySC is a mandatory dependency. I can make it optional, though I would find it misleading at this stage because there is no alternative solution and someone who wants to just correct the orbit will need to go look up for a special paragraph of the user guide.
> 
> Some points that I think are important to make:
> 
>     * The interfacing to pySC is a shortcut with little code and to me pySC does not need to be a permanent solution.
> 
>     * We can at any time decide that we don't want to use pySC and we can use something else, re-write pySC from scratch or even copy the relevant source code of pySC into pyAML.
> 
>     * Unlike having epics, tango, doocs, ophyd-async or bluesky as mandatory dependencies, pySC does not affect the design of the rest of pyAML.
> 
> 
> The biggest advantage I see now is that it allows us to develop how we want the configuration of the orbit correction to look like and how the user can interact with it, without locking into a specific technology.

I think this problem can be solved eventually by moving all applications to their own repositories and make the `pyaml` repository only include the middle layer parts that tie things together. Then we can follow the ecosystem approach I have seen be very useful for some other software stacks because it makes it easy for labs to pick and choose which parts they want to use. And also for people to add their own packages to the ecosystem and make them available for others to use if they want.

But for the prototype I think it makes sense to keep the applications in the same repository. It's already confusing for the standard matlab users who we want to get to test and provide feedback how to install the different repositories they need for their control system. 

For the installation instructions, @kparasch, what is the correct procedure for a user who wants to test the orbit correction in their machine? Does the pySC dependency have to be added to the pyproject.toml for them since they won't run the pytests? Or it's sufficient that they also do the step to update the submodules?

TeresiaOlsson:
@gubaidulinvadim I think the instructions are still a bit confusing for the users who want to test. I guess the point is that if they just want to test and not develop, they shouldn't install the pyaml repository at all but instead go to the package for the binding they need and install that? Is that correct?

JeanLucPons:
> @gubaidulinvadim I think the instructions are still a bit confusing for the users who want to test. I guess the point is that if they just want to test and not develop, they shouldn't install the pyaml repository at all but instead go to the package for the binding they need and install that? Is that correct?

Yes this is the idea.

gubaidulinvadim (approved):

PR #99, opened 2025-12-12
Accelerator load functions

JeanLucPons:
This PR is to discuss and implement `Accelerator` load functions.
I would prefer to keep the default `Accelerator.load()` for file but we can discuss.

### DIFF
--- a/pyaml/accelerator.py
+++ b/pyaml/accelerator.py
@@ -111,7 +111,7 @@ class Accelerator(object):
     @staticmethod
     def from_dict(config_dict: dict, ignore_external=False) -> "Accelerator":
         """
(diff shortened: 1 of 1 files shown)


gubaidulinvadim (dismissed):
I also prefer the default load() to have file as an input (yaml/json/toml)

gubaidulinvadim (approved):

TeresiaOlsson:
I imagined a solution where the load function understands itself what type to load depending on the input you give. So if you give it a filename it will use the loader for a file automatically. Then in the future if you instead give some database, web server etc it will use the loader for that. Then the loading interface for the user is the same independently of source.

But for now the only option will be a file anyway.

PR #102, opened 2025-12-15
Add ophyd example using BESSY VA.

JeanLucPons:
This PR add a file (EBSTune-ophyd.yaml) for testing pyaml_cs_oa backend


TeresiaOlsson:
I have added a page in the documentation where instructions for running this example (and the TANGO example) can go. I will try to test it later this week or in the beginning of next and add some instructions there for the workshop.

PR #110, opened 2025-12-17
WIP: some draft notes for developpers

simoneliuzzo:
added a notebook to describe how the development process works


TeresiaOlsson:
But pyaml-cs-oa will be required for the pyaml EPICS example? Otherwise there is only a TANGO example for pyaml.

For accml that's completely separate. That has it's own way to handle TANGO and EPICS that are separate from both tango-pyaml and pyaml-cs-oa.

JeanLucPons:
@TeresiaOlsson 
I also would like to try the bessy VA to see if i can connect with pyaml-cs-oa but it still fails:

```
[ubuntu20acu.pons] > apptainer run oras://registry.hzdr.de/digital-twins-for-accelerators/containers/pyat-softioc-digital-twin:v0-1-2-bessy.2475331
INFO:    Downloading oras image
255.8MiB / 255.8MiB [=======================================================================================================================================================] 100 % 43.7 MiB/s 0s
Environment variable MONGODB_URL is not defined, using default: mongodb://localhost:27017/bessyii
WARNING:dt4acc:Environment variable MONGODB_URL is not defined, using default: mongodb://localhost:27017/bessyii
Traceback (most recent call last):
  File "/twin/python/venv/bin/dt4acc_softioc.py", line 3, in <module>
    from dt4acc.custom_epics.ioc.server import main
  File "/twin/python/venv/lib/python3.12/site-packages/dt4acc/custom_epics/ioc/server.py", line 8, in <module>
    from .tasks import monitor_heartbeat
  File "/twin/python/venv/lib/python3.12/site-packages/dt4acc/custom_epics/ioc/tasks.py", line 9, in <module>
    ctx = ContextProxy("pva")
          ^^^^^^^^^^^^^^^^^^^
  File "/twin/python/venv/lib/python3.12/site-packages/dt4acc/custom_epics/utils/context_proxy.py", line 8, in __init__
    self.context = Context(context)
                   ^^^^^^^^^^^^^^^^
  File "/twin/python/venv/lib/python3.12/site-packages/p4p/client/asyncio.py", line 121, in __init__
    super(Context, self).__init__(provider, conf=conf, useenv=useenv, nt=nt, unwrap=unwrap)
  File "/twin/python/venv/lib/python3.12/site-packages/p4p/client/raw.py", line 154, in __init__
    self._ctxt = _ClientProvider(provider, conf=conf, useenv=useenv)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "src/p4p/_p4p.pyx", line 620, in p4p._p4p.ClientProvider.__init__
RuntimeError: Address family not supported by protocol
```

TeresiaOlsson:
> For the moment, i even cannot run the apptainer !? any idea ?

Hmm. Can you try to run my simple pyAT container? If that also doesn't work there is some more general problem with apptainer.

```
apptainer run oras://registry.hzdr.de/hzb/accelerator-tools/acc-containers/acc-container-pyat:latest
```

TeresiaOlsson:
> I managed to make it work. It create PV with a prefix having my user name:
> 
> ```
> epics tools ~ > caget pons:beam:twiss:x:tune
> pons:beam:twiss:x:tune         0.848009
> ```
> 
> I will try to set up a config file for it.

Automatically adding the username as a prefix is the way we handle being able to run several twins on the same network. So each user will need to have their own prefix in the config for the VA.

You want to use BESSY or MLS for the example? I was planning to go for MLS because smaller machine with less config to set up and individual power supplies for the quadrupoles so the tune example should already work.

JeanLucPons:
OK nice.
I will try with BESSY2 to set up a tune correction (using Q1 and Q2 family).

JeanLucPons:
Ok for Q3 and Q4.
I will update the management of serialized magnet (the simple part with one PS and one model).

JeanLucPons:
Thanks I think I have all what I need to set up a tune correction for bessy2.
I do that tomorrow, I hope i will not fall into troubles due to EPCIS.

TeresiaOlsson:
Okay. I need to think about that because the conversion factors should be the same ones that are used in MML so they should be compatible with pyAML. Unless there is a convention difference for the conversions between MML and pyAML? Do you have somewhere which numbers you used so I can double-check what they correspond to?

TeresiaOlsson:
Hmm, I'm confused now. The VA should not return K but current? And then pyAML should convert that into strength? And I thought that conversion between current and strength would be done in the same way as in MML and therefore use the same conversion factors?

JeanLucPons:
The VA returns well a current but the calcul using conversion factor yield to K and not to KL.
So for pyAML conversion factor need to be adjusted so design and live match.
I have a problem with Epics, the setpoint is not working. I'm close to finish the tune correction.

JeanLucPons:
pyAML use KL.

TeresiaOlsson:
> pyAML use KL.

I checked the MML manual and it uses K for the physics units. When was it decided that pyAML should use KL? If pyAML should change the convention I think a community discussion and decision is required because it will affect all existing MML users when transitioning to pyAML.

For us it doesn't matter so much because we anyway work in hardware units so I can easily change the conversion factors but for MML users that currently work in physics units it will be a major difference so they need to be involved in the decision. Maybe it's something to discuss at the workshop?

TeresiaOlsson:
Okay. I will make a discussion item to gather things we should discuss at the workshop. This seems to be one thing where a discussion might be required where different labs have different requirements that we need to satisfy somehow. Could be that pyAML for example natively works in KL but there must also be an option to use K in the same way as some labs need hardware units.

PR #117, opened 2026-01-12
Update docs to similar format as pyAT

TeresiaOlsson:
I have started to modify to get a similar structure as for the pyAT docs with a top menu for the main alternatives and then a left side menu for the submenus.

Definitely more work to do that I will continue with this week but I added a pull request in case someone else also has time to work on it or have input.


TeresiaOlsson:
I saw that in ophyd-async they use `autodoc2` instead of `autosummary` to generate the API documentation. ChatGPT claims that it might be a better option for large, complex projects. What do you think? Is it worth trying to see how it compares?

PR #122, opened 2026-01-14
New installation instructions

TeresiaOlsson:
I think our current installation instructions where the user is supposed to go to the repository for the bindings they want to use and install from there is too confusing. I think it makes it easier to understand if they are supposed to install the `pyaml` repository but with optional dependencies. Then there is the same entry point for installation independently of which control system you have.

I have updated the `pyproject.toml` to include that. For now it temporarily is installing directly from the github repositories.

I also added the steps to install like this to the docs.

What do you think? Is this a bad idea or you think it would help?

### DIFF
--- a/.github/workflows/deploy-pypi.yaml
+++ b/.github/workflows/deploy-pypi.yaml
@@ -17,6 +17,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
+      with:
+        submodules: recursive
     - name: Set up Python
       uses: actions/setup-python@v5
       with:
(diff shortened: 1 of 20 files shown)


docs/getting-started/installation.md
```diff
@@ -1,4 +1,28 @@
...
+
+        ```
+        pip install .[tango-pyaml]
+        ```
+
+    - **pyaml-cs-oa**: For TANGO or EPICS using `ophyd-async`.
+
+        ```
+        pip install .[pyaml-cs-oa]
+        ```
```
gubaidulinvadim: This is fine. I just note that before it was the inverse. One was installing tango-pyaml to get pyaml. This is also fine, but it is not good to put a git repository in pyproject.toml at this stage. The codes are on pip, it is better, in my view, to install a specific version that is known to be compatible.

PR #125, opened 2026-01-16
Chromaticity measurement alexandre moutardier

Amoutardier:
Module to measure chromaticity using frequency shift method.

### DIFF
--- a/tests/config/EBSTune.yaml
+++ b/tests/config/EBSTune.yaml
@@ -8,7 +8,7 @@ simulators:
     name: design
 controls:
   - type: tango.pyaml.controlsystem
-    tango_host: ebs-simu-2:10000
+    tango_host: ebs-simu-3:10000
     name: live
 data_folder: /data/store
 arrays:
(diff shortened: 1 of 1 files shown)


JeanLucPons:
Thanks @Amoutardier for this nice job !
Few remarks:

- All pyAML devices should be attached in post_init() to allow configuration error detection at early stage. `post_init()` is call after all underlying pyaml device are created.
Move to code below to `post_init()`.
```python
tune = self._peer.get_betatron_tune_monitor(self._cfg.betatron_tune)
rf = self._peer.get_rf_plant(self._cfg.RFfreq)
```

- Add the way to select the order and fit a polynomial rather than method=`linear` | `quad`, it you select 1 then `chromaticity_monitor.chromaticity.get()` should return $[[q'_x,q'_y]]$, if you select 2  $[[q'_x,q'_y] , [q''_x,q''_y]]$, etc...

- Add the way to retreive the tune scan data to allow the plot to de done by the user

PR #145, opened 2026-01-26
Update the BESSY II orbit example

TeresiaOlsson:
Updating the example for BESSY II orbit correction.

- Added more details in the README.
- Changed to new version of the twin.
- Updated BPM devices to use the PV available in the machine and not just the twin only PV.


JeanLucPons (approved):
Thanks @TeresiaOlsson for this thorough review !

simoneliuzzo:
@TeresiaOlsson 

this seems good to me! Thank you!

PR #151, opened 2026-01-28
Orm and dispersion attachement and callback

kparasch:
Attaches ORM and dispersion to element_holder so that you can do:
```
sr.design.orm.measure()
sr.design.dispersion.measure()
```

Also implements a handle to be able to run a callback function when measuring for user customization:
```
sr.design.orm.measure(callback=user_callback)
```
- [x] Attach ORM
- [x] Attach dispersion
- [x] Adapt ESRF example
- [x] Adapt BESSY example
- [x] Adapt tests
- [ ] Adapt documentation
- [x] Implement callbacks
- [x] Put example callback functions in the examples

### DIFF
--- a/examples/BESSY2_example/BESSY2Orbit.yaml
+++ b/examples/BESSY2_example/BESSY2Orbit.yaml
@@ -257,6 +257,17 @@ arrays:
       - BPMZ41D1R
       - BPMZ42D1R
 devices:
+- type: pyaml.tuning_tools.orbit_response_matrix
+  bpm_array_name: BPM
+  hcorr_array_name: HCorr
+  vcorr_array_name: VCorr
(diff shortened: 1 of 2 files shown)


kparasch:
I see, I understand your confusion. I put here the pseudo-code of running the response loop.

```
for corr in correctors:
    set k0 - delta
        -------> ACTION_APPLY
   get orbit
        -------> ACTION_MEASURE
    set k0 + delta
        -------> ACTION_APPLY
   get orbit
        -------> ACTION_MEASURE
    set k0
        -------> ACTION_RESTORE
```
In this case an iteration is complete each time ACTION_RESTORE is triggered

TeresiaOlsson:
No worries. I'm testing it right now.

PR #152, opened 2026-01-28
Updated example for BESSY II tune correction

TeresiaOlsson:
Updated example for the BESSY II tune correction.

Changes:
- Added more comments
- Added notebook
- Change config to use one quadrupole device per quadrupole to fit with BESSY II naming conventions and be able to have different conversion factors per magnet in the future and not just one per power supply.

This way of doing the config means the following:
- Response matrix for design and live doesn't agree -> this will be fixed later by using `SerializedMagnet`. A comment about that has been added to the example.

### DIFF
--- a/examples/BESSY2_example/bessy2-tune.ipynb
+++ b/examples/BESSY2_example/bessy2-tune.ipynb
@@ -13,7 +13,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "id": "1a092dbb",
    "metadata": {},
    "outputs": [],
@@ -23,7 +23,7 @@
     "import numpy as np\n",
     "\n",
     "from pyaml.accelerator import Accelerator\n",
-    "from pyaml.common.constants import ACTION_RESTORE\n",
+    "from pyaml.common.constants import ACTION_MEASURE\n",
     "from pyaml.magnet.magnet import Magnet"
    ]
   },
@@ -32,15 +32,7 @@
    "execution_count": null,
    "id": "9b63d253",
    "metadata": {},
-   "outputs": [
-    {
-     "name": "stdout",
-     "output_type": "stream",
-     "text": [
-      "28 Jan% 2026, 14:16:39 | WARNING | PyAML OA control system binding (0.1.1) initialized with name 'live' and prefix='a3744:'\n"
-     ]
-    }
-   ],
+   "outputs": [],
    "source": [
     "# ----- Load the configuration -----\n",
     "# Remember to change the prefix for the live mode to the one matching\n",
@@ -51,7 +43,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 3,
+   "execution_count": null,
    "id": "bf603b5f",
    "metadata": {},
    "outputs": [],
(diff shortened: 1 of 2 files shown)


TeresiaOlsson:
> @TeresiaOlsson Concerning how serialized magnet are implemented here:
> 
>     * When you have a list(), the total integrated strength is applied according the length of each magnet.
> 
>     * The calibration_factor are different from your config in accml as PyAML use the KL convention.
>       Basically each calibration factor are corrected by the magnet length.
> 
>     * If in your serie you have different magnets, I mean, for instance, different number of turn in their coils, then you need `SerializedMagnet` but i didn't tested it on a VA and it can be done later.
> 
> 
> Using the `lattice_names` field you can handle only uniform serialized magnets (same K), eventually with a different length.

- I saw this but didn't really like it. It worked but it's not the way our users would like to use the devices so I thought better to modify the example to match how we would like it to eventually look like for BESSY II so it can be used as a test case during the hackathon and later development.

- This I saw. I think I broke it when modifying the configuration but I will restore it now as part of benchmarking what currents I get between twin and MML so they agree.

- I don't think we have magnets with different number of coils in a series. Only the same magnet but they could potentially have a slightly different conversion factor since our conversion factors are beam-based extracted from LOCO. At the moment we have values fitted by D or T position but in the future we might want to fit by individual magnet instead. Could be that `SerializedMagnet` can already handle this. I will have to test it after the workshop.

The way you are doing it with the list might however work for MAX IV. It looked to me similar to how they currently have it in their MML since they do it differently compared to how we do it for BESSY II. In their MML config they have only one magnet device per power supply which I guess means their users are happy with using the device like that and don't need one device per magnet. After the workshop I can probably help @marlibgin2 and @xkillen to set up the config for R1 so they can try it.

TeresiaOlsson:
Hmm... I see what you mean. Then the linking for the simulator is not done by the power supply but using the lattice_names list. Kind of like if the lattice had a sliced magnet.

I will try it. Looks like it could solve the problem in the example that the response matrix measured on the design can't be used to correct on live.

TeresiaOlsson:
> If you want to keep the same name in pyaml than the lattice name nothing prevent to:
> 
> ```yaml
> - type: pyaml.magnet.quadrupole
>   name: Q3M1D2R
>   lattice_names: list(Q3M1D2R,Q3M2D2R)
>   model:
>     type: pyaml.magnet.linear_model
>     calibration_factor: -0.031715
>     unit: 1/m
>     powerconverter:
>       type: pyaml_cs_oa.epicsRW
>       write_pvname: Q3PD2R:set
>       read_pvname: Q3PD2R:rdbk
>       unit: A
> - type: pyaml.magnet.quadrupole
>   name: Q3M2D2R
>   lattice_names: list(Q3M1D2R,Q3M2D2R)
>   model:
>     type: pyaml.magnet.linear_model
>     calibration_factor: -0.0311902
>     unit: 1/m
>     powerconverter:
>       type: pyaml_cs_oa.epicsRW
>       write_pvname: Q3PD2R:set
>       read_pvname: Q3PD2R:rdbk
>       unit: A
> ```

I tested this but it makes it confusing what the strength means. I expect that the user wants to do something like this:
```
magnet = sr.design.get_magnet("Q3M2D1R")
strength = magnet.strength.get()
```

If I don't use the list that returns -5.0648 but if I use the list I get -5.0648*2. So it looks to the user that the strength of the Q3M2D1R magnet is twice what it should be. It's not just linking the elements together but also summing the strengths of all the magnets in the list.

I think that way of doing it only works for magnets that have been sliced in the lattice model but the users want to use it as one element in pyAML.

TeresiaOlsson:
This is ready to merge now.

PR #158, opened 2026-01-29
Doc build bindings

simoneliuzzo:
Pull request to try to make the documentation notebooks run correctly as in local builds

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -46,10 +46,10 @@ jobs:
 
       - name: Install package
         run: |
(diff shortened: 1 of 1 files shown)


TeresiaOlsson (dismissed):

TeresiaOlsson:
Maybe the problem is that `pip install accelerator-middle-layer[cs-oa-epics]` overrides the version of pyaml installed with `python -m pip install '.[doc]'`. I will see what happens if I try to install the bindings separately instead.

TeresiaOlsson:
Nope. Didn't fix it. Something else is the problem.

PR #162, opened 2026-01-30
ORM Tango server: added errors

JeanLucPons:
Added error to the lattice for better rendering


simoneliuzzo:
@JeanLucPons 

I do not understand clearly the plot with the dots. 
best regards
Simone

PR #166, opened 2026-01-30
Remove out of date notebook for how to run the BESSY II twin.

TeresiaOlsson:
I removed the notebook which is now out of date for how to run the BESSY II twin. The instructions are now only in the readme for the BESSY II examples.


kparasch (commented):
I didn't find the MacOs instructions to install the container in the readme of the bessy examples. Are they somewhere else?

TeresiaOlsson:
Looks like the apptainer webpage says:

limactl start template://apptainer

but the notebook also has:

limactl start –rosetta –vm-type=vz –network=vzNAT template://apptainer for ARM

limactl start –vm-type=vz –network=vzNAT template://apptainer for Intel

And then apptainer doesn't really explain in a good way how to use the limactl shell apptainer and not how to stop the container.

Maybe we should add those parts?

TeresiaOlsson:
I have added detailed instructions for how to install apptainer on different OS to the docs. @gubaidulinvadim @patrickmadela Maybe you also want to link to them in your example for the SOLEIL twin?

simoneliuzzo (approved):

PR #190, opened 2026-02-10
Updated code snippets

JeanLucPons:
Update code snippet for Arrays.

@gubaidulinvadim , @TeresiaOlsson , @simoneliuzzo 
It seems that I lost modifications I did previously regarding the big list of .rst files in [api](https://github.com/python-accelerator-middle-layer/pyaml/tree/main/docs/api) ?
Is this list automatically generated, in case it has nothing to do there or can we update it by hand ?


JeanLucPons:
@TeresiaOlsson 
The API reference is OK for me.
There are still duplicate warnings during the doc generation mainly link to wrong docstring in source.

JeanLucPons:
I ended the BPM array stuff.
Here is how it looks.
I can be used as a model to write code snippet, link, etc...

image

swhite2401 (approved):

PR #192, opened 2026-02-12
46 adding catalogs section in the pyaml configuration file

### LINKED ISSUE #187: Adding catalogs section in the pyaml configuration file
### Discussed in https://github.com/orgs/python-accelerator-middle-layer/discussions/45

<div type='discussions-op-text'>

<sup>Originally posted by **gupichon** February  5, 2026</sup>
## Proposal: Introducing a *Value Catalog* in pyAML

As discussed during the workshop, I would like to propose a new concept in **pyAML**: a **catalog of values**.

### Motivation

Today, control-system references (Tango attributes, strings, metadata, etc.) are spread across device and model definitions. This makes configurations harder to read, maintain, and reuse.

The proposed *value catalog* addresses this by centralizing and formalizing how such values are defined and accessed.

### Objectives

The main goals of this proposal are:

* **Centralize** all control-system references in a single, well-identified section of the configuration.
* **Expose a catalog of values** that can be accessed uniformly across pyAML.
* **Make these values usable as metadata**, enabling easy recording, logging, and export alongside measurements.

### Proposed Configuration Structure

With this approach, device and model definitions remain focused on physics and behavior, while all control-system bindings are moved to dedicated catalogs.

A possible configuration would look like this:

```yaml
type: pyaml.accelerator
facility: ESRF
machine: sr
energy: 6e9

simulators:
    lattice: sr/lattices/ebs.mat
    name: design

data_folder: /data/store

arrays:
  - type: pyaml.arrays.serialized_magnet
    name: series
    elements:
      - mySeriesOfMagnets

devices:
  - type: pyaml.magnet.serialized_magnet
    name: mySeriesOfMagnets
    function: B1
    elements:
      - QF8B-C04
      - QF8D-C04
      - QD5D-C04
      - QF6D-C04
      - QF4D-C04
    model:
      type: pyaml.magnet.linear_serialized_model
      calibration_factors: 1.00504
      calibration_offsets: 0.0
      unit: m-1
      curves: sr/magnet_models/quadcurve.yaml

  - type: pyaml.magnet.quadrupole
    name: QF1E-C04
    model:
      type: pyaml.magnet.linear_model
      calibration_factor: 1.00054
      crosstalk: 1.0
      curve:
        type: pyaml.configuration.csvcurve
        file: sr/magnet_models/QF1_strength.csv
      unit: 1/m

    name: SH1A-C01
    mapping:
      - [B0, SH1A-C01-H]
      - [A0, SH1A-C01-V]
      - [A1, SH1A-C01-SQ]
    model:
      multipoles: [B0, A0, A1]
      units: [rad, rad, m-1]
    use-physics: true
```

### Control-System Catalog

All control-system bindings are moved into a dedicated catalog:

```yaml
control-system-catalog:
  type: pyaml.common.catalog
  refs:
    - reference: mySeriesOfMagnets
      device:
        type: tango.pyaml.attribute
        attribute: srmag/ps-corr-sh1/c01-a-ch1/current
        unit: A

    - reference: QF1E-C04
      device:
        type: tango.pyaml.attribute
        attribute: srmag/vps-qf1/c04-e/current
        unit: A

    - reference: SH1A-C01-H
      device:
        type: tango.pyaml.attribute
        attribute: srmag/ps-corr-sh1/c01-a-ch1/strength
        unit: rad

    - reference: SH1A-C01-V
      device:
        type: tango.pyaml.attribute
        attribute: srmag/ps-corr-sh1/c01-a-ch2/strength
        unit: rad

    - reference: SH1A-C01-SQ
      device:
        type: tango.pyaml.attribute
        attribute: srmag/ps-corr-sh1/c01-a-ch3/strength
        unit: m-1
```

### User-Defined Data Catalog

The same mechanism can also be used for global values that are not tied to a specific pyaml model:

```yaml
user-defined-data:
  type: pyaml.common.catalog
  refs:
    - reference: machine-current
      device:
        type: tango.pyaml.attribute
        attribute: ans/ca/machinestatus/current
        unit: mA

    - reference: machine-mode
      device:
        type: tango.pyaml.string_attribute
        attribute: ans/ca/machinestatus/functionMode
        unit: ""
```

### Key Properties of the Catalog

* All catalog entries are identified by a **key** (`reference`).
* Any value can be retrieved **anywhere in pyAML** using:

  * its exact key, or
  * a regular expression (for bulk access and grouping).
* Catalog entries are first-class objects and can be:

  * recorded automatically as metadata,
  * reused by multiple models or tools,
  * queried independently of the device hierarchy.

### Open Design Questions

This proposal raises important points that need further discussion:

* How do models declare which catalog keys they depend on?
  In other words, how do we clearly define *which values a model expects* and *how it resolves them* from the catalog?
* Whether a validation step is required to ensure that all mandatory catalog entries are present, and at which stage this validation should occur.

</div>

gupichon:
## Description

All references to the control system are moved to a catalog section.

## Related Issue

- Closes #187 

- [ ] new feature: was implemented in the following way... because...
- [ ] bugfix: was implemented in the following way... because...
- [ ] ...

Describe the changes that had to be made to an existing functionality (if they were made)

- [ ] First change: reimplemented in the following way... because
- [ ] Second change: reimplemented in the following way... because
- [ ] ...

## Testing
 - [ ] first test
 - [ ] second test
 - ...

- [ ] New and existing unit tests pass locally
- [ ] Tests were added to prove that all features/changes are effective
- [ ] The code is commented where appropriate
- [ ] Any existing features are not broken (unless there is an explicit change to an existing functionality)


gupichon:
You can even go further and have the control system register itself in the catalog. Such a catalog could then check the Tango database to see whether a device exists, fetch its units, and so on. In that case, you would not even need to list the devices anymore.
However, this would reintroduce the naming convention at the device level again.

gupichon:
If it's ok for you I continue with the rest of pyaml.

JeanLucPons:
You already rewrote and tested the examples ?
I'll be off Tomorow and Monday.

gupichon:
> I don't understand why the name is needed there ?

Because the model does not know the name of the BPM.
But if the control system is responsible for retrieving the `DeviceAccess`, it is no longer needed.

gupichon:
It expects the following entries to be present in the catalog:

- `BPM_C05-01\x_pos`
- `BPM_C05-01\y_pos`
- `BPM_C05-02\x_pos`
- `BPM_C05-02\y_pos`

Otherwise, a `PyAMLException` will be raised: Catalog reference '{reference}' not found.

gupichon:
It is done.

JeanLucPons:
It fails with BPM aggregator.
```python
            for b in bpms:
                devs = self.attach_array(b.model.get_pos_devices())
                devH = devs[0]
```
I'm investigating, i observed a strange issue.

JeanLucPons:
I agree.
I did that a bit quickly before the workshop to make the BESSY2 example working.
Aggregators need to be refurbished and indexed attribute concept generalized.
I'll test the BESSY2 example today.

gupichon:
Ok, those problems already existed without the catalog. If you’re OK with it, I’ll let you deal with them. I suggest we continue focusing on the catalog topic. As @TeresiaOlsson mentioned [here](https://github.com/orgs/python-accelerator-middle-layer/discussions/47#discussioncomment-15858166), we need to test that a change to a reference in the catalog is correctly propagated.

JeanLucPons:
Yes, yes, the only thing here, is to make BESSY example work with this Catalog.
Basically this file:
https://github.com/python-accelerator-middle-layer/pyaml/blob/46-adding-catalogs-section-in-the-pyaml-configuration-file/examples/BESSY2_example/BESSY2Orbit.yaml
I let you deal with Soleil example

gupichon:
@TeresiaOlsson, as requested, changing the device associated with a reference in the catalog will take effect everywhere.
@JeanLucPons, we should discuss the ControlSystem class in PyAML. The way the attach methods are currently used requires refactoring.

JeanLucPons:
I do not see the point of having dynamic references to static stuff in a config file !
Unless you implement dynamic catalog.

JeanLucPons:
To go a bit further, and to avoid confusion, it would even change property to *_name:

```python
sr.live.get_bpms("BPM").x_pos_name = "ORBITCC:rdPos"
or
sr.live.get_bpms("BPM").x_pos_name = ["ORBITCC:rdPos"] * len(sr.live.get_bpms("BPM"))

sr.live.get_bpms("BPM").x_pos_index = [i*2 for i in range(len(sr.live.get_bpms("BPM"))]
sr.live.get_bpms("BPM").y_pos_index = [i*2+1 for i in range(len(sr.live.get_bpms("BPM"))]
```

TeresiaOlsson:
They are also really thinking about it like it's a database despite that it in some cases is just a json file. Here are some examples of the user interface: https://pcdshub.github.io/happi/v3.0.1/client.html#db-choice

I quite like that interface because it means that the yaml file is no longer a long confusing text file but just the backend for a database that the user isn't meant to look at directly. Instead they would browse, add and modify things in it through a python API. And that same Python API can also be used to create it from scratch for new users. Maybe that would solve our problem where every lab now need to write their own scripts to generate the yaml file? Now it's a bit like everyone needs to invent that API by themselves.

gupichon:
Thanks a lot, @TeresiaOlsson, this will be very helpful. I’m going to restart this development from scratch.

JeanLucPons:
You want to restart this PR from scratch ? Using happi ?
To me, after a very first glance, it seems that it could replace pydantic, if happi handle validation ?

TeresiaOlsson:
I'm also exploring to make a similar implementation as in HAPPI because it seems like the way to go. I played with it a bit last week and started to like that approach more and more.

But I would not use HAPPI directly because it feels like a dependency we don't want to have. I don't think there is a big community around that package so it might not be long-term sustainable to depend on. I also think HAPPI doesn't do validation and some of the things they do could have been implemented easier if pydantic basemodels were used instead. The happyItem and containers look to me pretty much like they could have been basemodels.

TeresiaOlsson:
@gupichon I pushed what I had been playing with so far into the branch [happi-style-config](https://github.com/python-accelerator-middle-layer/pyaml/tree/happi-style-config). So far I only tried to do a very basic yaml and json backend to understand how that might work compared to how we use the config file now. No guarantee that it actually is working because I didn't come so far.

But at least my feeling was that we like this could have pretty much the same yaml file format as now.

gubaidulinvadim:
@TeresiaOlsson @JeanLucPons @gupichon I guess I'll put this discussion on the agenda for the meeting on Friday.

TeresiaOlsson:
I will be on holiday on Friday so I can't come. But if pyAML decides to do it in some other way I will keep developing it as an external package which would allow the EPICS labs to have some kind of device database that we can use to generate the config file for pyAML. At the moment it's just way too difficult for us since we don't have the TANGO database that we can extract information from. I think some EPICS labs might already have something like this but at HZB we currently only store this information inside the MML config which isn't great so we need to replace it.

JeanLucPons:
- set_energy() function has been moved to ElementHolder, so should be removed from control_system.py.

JeanLucPons:
What is this the purpose of this unwrap method in DeviceAccessList ?
```
  This proxy behaves like the wrapped DeviceAccess:
    - Unknown attribute access is forwarded to the current target (e.g. target._cfg).
    - Unknown attribute assignment is also forwarded to the target.
    - The proxy remains a stable handle whose target can be swapped at runtime.
```
The above is unclear for me and IMHO this should be handle in the ControlSystem backend.
I would like to have a single string to access an attribute.
If the string changes (via a validation decorator or an other config wrapper) then the PyAML Element should trigger the backend to renew the DeviceAccess.
i.e.:
```python
sr.live.get_betatron_tune_monitor("BETATRON_TUNE")._cfg.tune_h = "something" # Incorrect
# should be replaced by
sr.live.get_betatron_tune_monitor("BETATRON_TUNE").tune_h = "something" # Trigger a creation of DeviceAccess
sr.design.get_betatron_tune_monitor("BETATRON_TUNE").tune_h # should throw an exepction
```

gupichon:
Yes, anyway, I’m going to abandon this branch and redo the catalog proposal from scratch. I will use this branch as a reference for some parts of the implementation and for the discussions we had.
After discussing it with @lnadolski, I will also propose managing the configuration as a service and building the PyAML objects from that service. Possibly with a DTO pattern, but I’m not sure yet.
The idea is to make the configuration accessible without PyAML itself. For example, this could allow access to the list of quadrupoles used for tune control.

JeanLucPons:
Ok.
It would be nice to address this during the next maintainers meeting.
I would like also to write a GUI (jive like) may be on top of the service you propose.

gupichon:
> Ok. It would be nice to address this during the next maintainers meeting. I would like also to write a GUI (jive like) may be on top of the service you propose.

Such a GUI make complete sense, it would be wonderful. Since the next meeting, I think I will have time to write a complete proposal into an issue. I'm working on it. The idea is to be able to compose a configuration from several parts easily but in the end being able to keep the actual behaviour with the yaml file for the user point of view.

TeresiaOlsson:
Would this allow us to have a separate configuration layer which only contains data and no business logic? If so that would be great because I think the current architecture where the same objects are both responsible for storing the configuration data and doing the business logic will be difficult to maintain long-term.

JeanLucPons:
I would say yes.
The idea is to be able to create pyaml yaml/json files from an advanced tool.
Then pyaml will be able to load and create its objects from a either a dict, a yaml or a json file.

gupichon:
Yes, we will discuss the final implementation elsewhere, but the idea is to provide PyAML with what it needs to use the service (e.g., the file path), and then it will query the service to build its objects.
From the user’s point of view, the existing call will remain available, but other ways to build the configuration will also be introduced, offering more options and a more accessible API.

gubaidulinvadim:
Continues with new implementation in https://github.com/python-accelerator-middle-layer/pyaml/pull/234

PR #213, opened 2026-03-09
Change to using automatic versioning from Git tags

TeresiaOlsson:
I have added automatic versioning from Git tags in the pyproject.toml. Advantages of this:

- There is no hardcoded version number that needs to be manually bumped when making a new release. Instead the release can be built based on a Git tag which simplifies the release workflow.

- If you install directly from the Git repository you will get a version number based on the next release number + the commit number making it easier to distinguish between the developer version and the released version. This helps to know which version you have installed when debugging. For example it might look like this in `pip list`:

```
accelerator-middle-layer  0.2.4.dev166+g983a827d1.d20260309
```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -28,16 +28,8 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install numpy
-        pip install scipy
-        pip install pydantic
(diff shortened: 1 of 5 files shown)


gupichon (dismissed):
Seems ok to me

gupichon (approved):
Tested locally. Works fine.

PR #214, opened 2026-03-10
Add a validation decorator

TeresiaOlsson:
I have started an idea to introduce a validator decorator that you can use to specify which ConfigModel to use to validate a specific class. The idea is:

- Instead of enforcing that a module needs to contain a main class defined by the `PYAMLCLASS` and a configuration model that has to be named `ConfigModel` use a decorator to attach a validation model to a specific class. For example like:

  ```
  @validator(ConfigModel)
  class Accelerator(object):
  ```
  
    This would add an attribute `_validator_model` but also two functions `validate` and `from_validated`. `validate` allows you to validate the input data without creating an object of the class (for example if you want to create the dict separately but want some way to check that it's okay before loading it). `from_validated` validates the input data and creates the object.

- This would make it easier to integrate devices/applications from third-party packages where it's not possible to enforce a certain module structure. As long as a class has the attributes/methods required by pyAML I think it should be possible to integrate and configure it in the same way as the standard pyAML classes despite that the module has been written differently.

- The docs are currently difficult to understand because you can't see which attributes a class needs without having to also look at the config class. Instead I suggest to not build the config model into the class but separate validation and object creation.

  This means you need to write the required attributes twice which is a bit annoying, but at the same time it makes it possible to create objects without mandatory validation (which could be useful if you already have a validated config and don't need to validate every time) and make the dependency of pydantic lighter.

I have so far made a test only including the `Accelerator` to show the idea and try how it can be implemented. Before I continue I would like input on what you think about the idea and my suggestion for how to implement it?

- I know the field_locations are currently not included. I'm working on getting them back.
- It will also not yet work for building any other objects than the accelerator. For testing only a minimal config will work so far:

    ```
    type: pyaml.accelerator
    facility: BESSY2
    machine: storage_ring
    energy: 1.7e9
    ```

### DIFF
--- a/pyaml/accelerator.py
+++ b/pyaml/accelerator.py
@@ -3,6 +3,7 @@ Accelerator class
 """
 
 import os
+from typing import Optional
 
 from pydantic import BaseModel, ConfigDict, Field
 
@@ -11,6 +12,7 @@ from .common.element import Element
 from .common.exception import PyAMLConfigException
 from .configuration.factory import Factory
 from .configuration.fileloader import load, set_root_folder
+from .configuration.validation import validator
 from .control.controlsystem import ControlSystem
 from .lattice.simulator import Simulator
 
@@ -36,13 +38,13 @@ class ConfigModel(BaseModel):
         can access several control systems
     simulators : list[Simulator], optional
         Simulator list
-    data_folder : str
+    data_folder : str, optional
         Data folder
     arrays : list[ArrayConfig], optional
         Element family
     description : str , optional
         Acceleration description
-    devices : list[.common.element.Element]
+    devices : list[Element], optional
         Element list
     """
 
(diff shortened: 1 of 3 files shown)


JeanLucPons:
OK
If I understand well your idea, you would like to have a type field (in the config) that refer to a class and not to a module ?
The doc is structured in a way that the ConfigModel appears at the beginning and no duplication are needed or I miss something ?

Please do not remove the method `def handle_validation_error()` that creates a comprehensible pydantic error messages in union with other error detection message in the yaml file. Or at least take care of providing good error messages.

TeresiaOlsson:
> I have nothing against removing the `PYAMLCLASS` and having a class in the type field instead of a module but this may prevent to use introspection to generate doc, dynamic API, etc... We can also support both a `type` field that refer to a module as today and a `class` field that refer to a random class (with or without pydantic validation) and let the user choose between one of the 2. Edit: In any case, we need a mechanism to validate interface of measurement or tuning tools so a user has to connect something compatible. For instance, a Tune monitor object should have an attribute tune `ReadFloatArray` that returns [qx,qy].

I like the idea of supporting a `type` field and a `class` field. I can explore options for that after I have finished the work with this decorator for the validation.

For validating the interface I think the easiest is to use protocols. Then if a third-party application doesn't already fulfill them it doesn't feel so complicated to me to write a little wrapper that imports the protocols and adds the missing functionality. But maybe there are other options also.

TeresiaOlsson:
I will remove this PR since the idea has been replaced with the idea of a SchemaRegistry in PR #264.

Issue #155, opened 2026-01-28
Tests need internet connection to run


kparasch:
I realized during my airplane flight that the tests are failing if you don't have internet connection.

The issue seems to be related to the fact that we are pip-installing the dummy-cs/tango-pyaml bindings, even though we are using a local folder.

gupichon:
It should be fine with the latest release. I’ve tested it on my end, but I’d feel more comfortable if someone else could verify it as well.
@kparasch ?
@JeanLucPons ?
@gubaidulinvadim ?

gubaidulinvadim:
Latest release or latest commit?

gupichon:
commit ! last commit ! Sorry...

gubaidulinvadim:
It works.

PR #231, opened 2026-04-16
228 feature allow incremental accelerator configuration with config manager

### LINKED ISSUE #228: Feature: allow incremental Accelerator configuration with add_config()

`Accelerator` is currently initialized from a full YAML/JSON configuration file via `Accelerator.load(...)`, or from a full configuration dictionary via `Accelerator.from_dict(...)`.

In practice this makes simple programmatic use cases unnecessarily heavy. Even for small setups, users need to maintain a full YAML file (or set of files) such as:

```python
acc = Accelerator.load("tests/config/bpms.yaml")
```

This works well for static file-based configurations, but it is not convenient when the configuration is naturally assembled in Python from multiple partial dictionaries, or when part of the configuration is provided by local or remote resources.

There is currently no supported way to:
- create an empty `Accelerator`
- progressively add configuration fragments through a single public API
- fetch raw configuration content from a local file or a REST endpoint without immediately building an `Accelerator`

**Proposed solution**

Introduce a programmatic builder-style API for `Accelerator`, centered on a single method for configuration assembly:

- add `Accelerator.empty()` or `Accelerator.builder()`
- add a single method `add_config(...)` to merge configuration fragments into the current accelerator-under-construction
- add a mirror method `remove_config(...)` to remove previously attached configuration fragments from the current accelerator
- add a static method `Accelerator.fetch_config(...)` to retrieve a raw configuration dictionary from either a local file or a remote URL

Expected behavior:
- `add_config(...)` accepts a dictionary fragment, a local YAML/JSON file path, or a remote URL returning a dictionary payload
- `add_config(...)` may optionally accept a `key` argument to extract a specific fragment from the loaded payload before merging it
- each call to `add_config(...)` validates the merged configuration immediately, so configuration errors are detected as early as possible
- each call to `add_config(...)` updates the live `Accelerator` object immediately, so newly added capabilities become available right away
- if an `add_config(...)` operation fails after partially updating list-based sections such as `devices`, `controls`, or `simulators`, the accelerator state must remain consistent and the added entries must be removable or rolled back cleanly
- `remove_config(...)` removes a previously attached configuration fragment and updates the live `Accelerator` object immediately
- each call to `remove_config(...)` validates the resulting configuration immediately after removal
- `fetch_config(...)` returns the loaded dictionary or selected fragment without building PyAML objects and without merging anything
- existing `Accelerator.load()` and `Accelerator.from_dict()` behavior remains backward compatible

One alternative would be to keep the current `Accelerator.from_dict()` workflow and ask users to manually fetch remote content, load local files themselves, merge all fragments externally, and only then call `from_dict()`.

This is workable, but it pushes repetitive boilerplate to users and makes common programmatic use cases unnecessarily cumbersome.

Another alternative would be to add multiple specialized methods such as `add_dict(...)`, `add_file(...)`, and `add_url(...)`. This does not seem preferable because it fragments the public API, while a single `add_config(...)` entry point would be easier to understand and use.

Similarly, relying only on implicit rollback inside `add_config(...)` would not fully address operational needs. In practice, users may also need an explicit mirror operation to remove a previously added named fragment from the accelerator state.

**Example**
```python
acc = Accelerator.empty()

acc.add_config({
    "facility": "ESRF",
    "machine": "Test",
    "energy": 6e9,
    "data_folder": ".",
})

acc.add_config({
    "devices": [
        {...},
        {...},
    ]
})

acc.add_config("tests/config/devices.yaml")

acc.add_config("http://config-service/api/config", key="controls.live_shadow")

acc.remove_config({"controls": [{"name": "live_shadow"}]})
```

Raw fetching would also be available independently:

```python
cfg = Accelerator.fetch_config("tests/config/devices.yaml")
fragment = Accelerator.fetch_config(
    "http://config-service/api/config",
    key="controls.live_shadow",
)
```

**Additional context**

This feature would make `Accelerator` easier to use from tests, notebooks, and higher-level tools that assemble configuration dynamically.

It would also help for integration with lightweight REST configuration services. A typical use case would be a generic endpoint such as `GET /api/config` returning several sections, from which PyAML could optionally extract only one fragment such as `controls.live_shadow`.

There is also an important scope limitation for partial loading. Deep configuration fragments should not be expected to load correctly in isolation when they require descriptive context or when they only make sense once attached to an existing PyAML object graph. For example, magnet conversion models are typically not meaningful as standalone fragments.

In practice, partial loading should therefore be limited to named top-level or attachable elements such as tools, magnets, BPMs, control systems, simulators, arrays, or similar objects that can be unambiguously merged into the accelerator configuration. Loading arbitrary deep subparts of the configuration without context should not be a goal of this feature.

If support for deep object-level updates is needed later, that should probably be handled in a separate issue. A possible direction would be to explicitly mark some fields as remotely or incrementally addressable, for example through a decorator or a similar declaration mechanism on selected fields such as `DeviceAccess`. This is a different problem from the incremental addition and removal of named PyAML objects and should remain out of scope here.

Another important point is error handling during incremental updates. Some `add_config(...)` calls will enrich list-based sections such as `devices`, `controls`, `simulators`, or arrays. If validation or attachment fails after some of these entries were already added, the operation should not leave the accelerator in a partially mutated state. The implementation should therefore support clean rollback, or an equivalent remove/revert mechanism for the newly added entries.

Beyond rollback on failure, there should also be an explicit mirror API for removal. This is especially relevant for named elements stored in list-based sections, where users may want to detach a previously added control system, simulator, tool, BPM, magnet, or similar object from the current accelerator configuration.

Suggested acceptance criteria:
- users can initialize an accelerator without an existing full local YAML/JSON file
- users can progressively assemble configuration using only `add_config(...)`
- users can explicitly remove previously attached configuration using `remove_config(...)`
- `add_config(...)` accepts dictionaries, local file paths, and URLs
- users can optionally provide a key to select a specific fragment from the loaded payload
- each `add_config(...)` call validates the resulting merged configuration automatically
- each `remove_config(...)` call validates the resulting configuration automatically
- failed `add_config(...)` operations do not leave partially added entries behind in list-based sections
- partial loading is limited to named elements that can be attached to the accelerator configuration; unsupported deep fragments fail explicitly
- `Accelerator.fetch_config(...)` returns the raw loaded dictionary or selected fragment without building objects
- tests cover incremental construction, local file loading, REST-based loading, keyed fragment extraction, and raw fetching via `fetch_config(...)`

**Checklist**

gupichon:
## Description

Add a `ConfigurationManager` that can aggregate accelerator-root YAML/JSON/dict fragments before building an `Accelerator`. This makes incremental configuration assembly explicit, keeps `Accelerator.load()` compatible for full accelerator configs, exposes inspection/query helpers for merged configuration state, and now supports REST-backed YAML/JSON sources with relative includes resolved from the base URL.

## Related Issue

This project only accepts pull requests related to open issues. If suggesting a new feature or change, please discuss it in an issue first.

- Fixes #228

- [x] new feature: introduced `pyaml.configuration.ConfigurationManager` to add, inspect, query, replace, remove, clear, snapshot and build accelerator configuration fragments incrementally because the previous flow only supported loading a single full accelerator file.
- [x] bugfix: updated `Accelerator.load()` to validate that the root config is `pyaml.accelerator` and raise a clearer error for sub-element roots because the new configuration flow now distinguishes full accelerator builds from direct factory usage.
- [x] new feature: documented and exported `ConfigurationManager` in `pyaml.configuration` and generated API docs, including exploratory special members, because the new API is intended for direct public use.
- [x] new feature: added REST configuration loading through a dedicated fetcher module, including recursive remote YAML/JSON includes and relative path resolution from the remote base URL, because configuration fragments may now be hosted remotely instead of only on the local filesystem.

Describe the changes that had to be made to an existing functionality (if they were made)

- [x] `Accelerator.load()` now builds through `ConfigurationManager.add(...).build(...)` instead of calling the file loader directly, while preserving support for standard accelerator files.
- [x] `Accelerator.load()` no longer assumes a local filesystem path up front, so full accelerator configurations can now be loaded from local files or `http`/`https` URLs through the same manager-based flow.
- [x] `ConfigurationManager` now delegates remote fetching/parsing to `pyaml.configuration.restfetcher` rather than embedding HTTP logic in `manager.py`, keeping network concerns isolated from merge logic.
- [x] Test fixtures no longer install temporary test packages with `pip`; they now expose package roots through `sys.path` and purge imported modules between tests, which avoids editable/install side effects during the test suite.
- [x] Several tests that previously loaded monolithic YAML files now build accelerators from multiple fragments, proving the incremental configuration path on arrays, devices and tune-monitor use cases.

## Testing
 - [x] `tests/test_configuration_manager.py`
 - [x] `tests/test_accelerator_load.py`
 - [x] Extended regression coverage in `tests/test_arrays_ops.py`, `tests/test_errors.py`, `tests/test_factory.py` and `tests/test_tune_monitor.py`
 - [x] Added local HTTP-server fixtures in `tests/conftest.py` to exercise remote configuration loading and base-URL include resolution
 - [x] Verified locally with `pytest tests/test_configuration_manager.py tests/test_accelerator_load.py` (`17 passed`)

- [x] Any existing features are not broken (unless there is an explicit change to an existing functionality)

### DIFF
--- a/tests/test_errors.py
+++ b/tests/test_errors.py
@@ -3,6 +3,7 @@ import pytest
 from pyaml.accelerator import Accelerator
 from pyaml.arrays.magnet_array import MagnetArray
 from pyaml.common.exception import PyAMLConfigException, PyAMLException
+from pyaml.configuration import ConfigurationManager
 
 
 @pytest.mark.parametrize(
@@ -23,6 +24,7 @@ def test_tune(install_test_package):
         ml: Accelerator = Accelerator.load("tests/config/bad_conf_duplicate_3.yaml")
     assert "Configuration entry 'BPM_C04-06' is duplicated inside category 'devices'" in str(exc)
     assert "bad_conf_duplicate_3.yaml" in str(exc)
+    assert "line 58, column 3" in str(exc)
 
     with pytest.raises(PyAMLConfigException) as exc:
         ml: Accelerator = Accelerator.load("tests/config/bad_conf_duplicate_4.yaml")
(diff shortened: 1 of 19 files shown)


tests/test_errors.py
JeanLucPons, line 1: Why the line number information is lost here ?
Is there a way to restore it ?

JeanLucPons:
Test are failing now.
It is due to bad install of dummy-cs.
You did lots of mods in conftest.py and now it seems that it fails in installing dummy-cs.

image
image

gupichon:
That's strange. It's working on my end. I'll try with a fresh environment.

gupichon:
@JeanLucPons, that’s very strange. These are exactly the issues I observed on GitHub and solved yesterday. Locally, I have no problems at all. I rebuilt the entire environment and everything is working.

JeanLucPons:
The issue is reproductible.
If you have tango-pyaml installed, the test fail.
This is not happening in main.
So it seems that there are issue in conftest.py where the dummy-cs is not correctly installed when tango-pyaml is already installed.

gupichon:
I tried, and it is indeed the GitHub use case. I'm still trying to reproduce it at the office. Apparently, it’s only happening locally on Ubuntu, which is even weirder!
I'll fix it. The tests are running well on github, so I think it should be quick to fix once I can reproduce it.

gupichon:
So, I tested it on Windows and Ubuntu (via WSL) and GitHub Actions passed. I tried both with and without installing dummy-cs beforehand.

@JeanLucPons, I’ll let you update and test again, but I’d like to understand how it could fail again.

By the way, the update to conftest is to prevent the use of pip install, which requires an internet connection and is much slower.

I'll look into another optimization in a separate issue: caching the lattice for tests only. It reduces test execution time by a factor of 10, but it could introduce some hidden bugs...

JeanLucPons:
I think the problem comes when tango-pyaml is already installed. the dummy-cs is not forced.

gupichon:
I recommend using two different environments. It’s much easier and it’s the proper way to handle it, as PyAML tests will fail if tango-pyaml is installed. This is actually expected behavior.

JeanLucPons:
OK what about the "Error line number" information lost ?

gubaidulinvadim:
> By the way, the update to conftest is to prevent the use of pip install, which requires an internet connection and is much slower.

Please, close the issue #155 if that is the case

gupichon:
> OK what about the "Error line number" information lost ?

It's back.

gupichon:
> OK what about the "Error line number" information lost ?

It's back and the tests have been improved. The new difficulty lay in handling multiple possible sources and the different potential errors.

JeanLucPons:
So OK for me

JeanLucPons (approved):

PR #232, opened 2026-04-17
Interface for tune monitor + mods for tune in frequency

JeanLucPons:
This PR add an abstract interface (similar to a Protocol) to the BetatronTuneMonitor in view of adding external tune monitor.

### DIFF
--- a/pyaml/accelerator.py
+++ b/pyaml/accelerator.py
@@ -35,6 +35,8 @@ class ConfigModel(BaseModel):
         this value can be dynamically set
     alphac : float, optional
         Moment compaction factor.
+    harmonic_number: int, optional
+        Number of bucket
     controls : list[ControlSystem], optional
         List of control system used. An accelerator
         can access several control systems
@@ -56,6 +58,7 @@ class ConfigModel(BaseModel):
     machine: str
     energy: float
     alphac: float | None = None
+    harmonic_number: int | None = None
     controls: list[ControlSystem] = None
     simulators: list[Simulator] = None
     data_folder: str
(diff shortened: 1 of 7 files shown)


pyaml/diagnostics/tune_monitor.py
```diff
@@ -64,6 +71,36 @@ def tune(self) -> ReadFloatArray:
...
+        class TuneFreq(ReadFloatArray):
+            def __init__(self, parent: BetatronTuneMonitor):
+                self.parent = parent
+
+            def get(self) -> np.array:
+                h = self.parent._h
+                rf_name = self.parent._cfg.rf_plant_name
+                if h is not None and rf_name is not None:
+                    tune = self.parent.tune.get()
+                    rf = self.parent._peer.get_rf_plant(rf_name)
```
kparasch, line 94: I wonder if it would be simpler to use "design" revolution frequency instead of rf_frequency/harmonic_number.
rf_frequency (and revolution frequency too) can change e.g. during chromaticity measurement, though the change will most likely be insignificant.

I think it is still ok like this.

pyaml/accelerator.py
```diff
@@ -56,6 +58,7 @@ class ConfigModel(BaseModel):
     machine: str
     energy: float
     alphac: float | None = None
+    harmonic_number: int | None = None
```
kparasch, line 61: Should we create something like an "optics parameters" container for variables like this? I suspect we will need to store more and more as we develop?

pyaml/diagnostics/tune_monitor.py
```diff
@@ -64,6 +71,36 @@ def tune(self) -> ReadFloatArray:
...
+                    freq = rf.frequency.get()
```
kparasch, line 95: Will rf.frequency always be defined in Hz?

kparasch (approved):

pyaml/diagnostics/tune_monitor.py
JeanLucPons: Yes
Unit conversion shoud be handle in the rf plant itself.
At the pyaml level we expect SI.

pyaml/accelerator.py
JeanLucPons: Why not. this can be easily done later

PR #233, opened 2026-04-17
Allow Ophyd devices

### LINKED ISSUE #230: Feature: Ophyd-async device integration
A way to integrate our existing ophyd devices for IDs that we use to measure feed forward tables.

 _Originally posted by @TeresiaOlsson in [#228](https://github.com/python-accelerator-middle-layer/pyaml/issues/228#issuecomment-4251394289)_

**Proposed solution**
A module binding OphydDevice (meaning ophyd-async) to PyAML device.
...

One possible alternative could be to have ophyd-cs-oa to bind directly OphydDevice (ophyd-async) instead of OphydSignal (?). 
...

**Example**
...

**Additional context**
One example of where this is needed is the interaction between the operation and insertion device (ID) groups at HZB. ID group already has existing ophyd-async devices that should be used/orchestrated together with pyAML measurements if one wants to gather data for ID feedforward tables.

**Checklist**

JeanLucPons:
This PR adds a way to define an external tune monitor and a way to define object using class name without implicit model validation. Each `ControlSystem` element holder is populated by such elements according to `control_modes` property.

Example:

```python
from pyaml.common.element import Element
from pyaml.diagnostics.atune_monitor import ABetatronTuneMonitor
from pyaml.common.abstract import ReadFloatArray
from pyaml.control.controlsystem import ControlSystemAdapter
import numpy as np
from pydantic import BaseModel
from typing import TypedDict
from ophyd_async.core import StandardReadable, Transform, DerivedSignalFactory
from ophyd_async.epics.core import epics_signal_r
import asyncio

evloop = None

class MyControlSystemConfigModel(BaseModel):
    name: str
    prefix: str
    view: str

class MyControlSystem(ControlSystemAdapter):
    def __init__(self, cfg: MyControlSystemConfigModel):
        ControlSystemAdapter.__init__(self)
        self._cfg = cfg
    def name(self) -> str:
        return self._cfg.name
    def prefix(self) -> str:
        return self._cfg.prefix
    def view(self) -> str:
        return self._cfg.view
    

class TuneDerived(TypedDict):
    frequency: float
    tune: float
    
class TuneTransform(Transform):
    revolution_frequency: float
    hardware_view: bool

    def raw_to_derived(self, *, signal: float) -> TuneDerived:
        if self.hardware_view:
            frequency = signal * 1e3
            tune = frequency / self.revolution_frequency
        else:
            tune = signal
            frequency = tune * self.revolution_frequency * 1e3

        return TuneDerived(frequency=frequency, tune=tune)
    

class Tune1D(StandardReadable):
    """Device to read tune in one dimension."""

    def __init__(
        self,
        peer:"MyTuneMonitor",
        cs:"MyControlSystem",
        name: str,
        ):

        native = epics_signal_r(float, name)
        self._factory = DerivedSignalFactory(
                TuneTransform,
                signal=native,
                revolution_frequency=peer.frev,
                hardware_view=cs.view() == "HARDWARE"
        )

        if cs.view() == "HARDWARE":
            frequency = native
            tune = self._factory.derived_signal_r(float, "tune")
        else:
            tune = native
            frequency = self._factory.derived_signal_r(float, "frequency")

        with self.add_children_as_readables():
            self.tune = tune
            self.frequency = frequency

        super().__init__(name=name)  


class MyTuneMonitor(Element,ABetatronTuneMonitor,StandardReadable):

    def __init__(self,name:str,cs:MyControlSystem,device_h:str,device_v:str,frev:float):

        Element.__init__(self,name)
        self.frev = frev
        tune_h = Tune1D(self,cs,cs.prefix()+device_h)
        tune_v = Tune1D(self,cs,cs.prefix()+device_v)
        with self.add_children_as_readables():
            self.hor = tune_h
            self.ver = tune_v
        StandardReadable.__init__(self,name=name)

    def arun(self,coro):
        return evloop.run_until_complete(coro)
    
    async def read_tune(self):
        await self.connect()
        req = [
          self.hor.tune.get_value(),
          self.ver.tune.get_value()
        ]
        return await asyncio.gather(*req)

    async def read_tunefreq(self):
        await self.connect()
        req = [
          self.hor.frequency.get_value(),
          self.ver.frequency.get_value()
        ]
        return await asyncio.gather(*req)
        
    @property
    def frequency(self) -> ReadFloatArray:
        class TuneFreqReader(ReadFloatArray):
            def __init__(self, parent: MyTuneMonitor):
                self.parent = parent
            def get(self) -> np.array:
                return self.parent.arun(self.parent.read_tunefreq())
            def unit(self) -> str:
                return "Hz"
        return TuneFreqReader(self)
    
    @property
    def tune(self) -> ReadFloatArray:
        class TuneReader(ReadFloatArray):
            def __init__(self, parent: MyTuneMonitor):
                self.parent = parent
            def get(self) -> np.array:
                return self.parent.arun(self.parent.read_tune())
            def unit(self) -> str:
                return "1"
        return TuneReader(self)



acc_config = {
    "type": "pyaml.accelerator",
    "facility": "BESSY2",
    "machine": "sr",
    "energy": 1.7185e9,
    "data_folder": "/data/store",
    "controls":
     [{"type": MyControlSystem.__module__,
       "class": "MyControlSystem",
       "validation_class": "MyControlSystemConfigModel",
       "prefix": "pons:",
       "name": "live",
       "view": "HARDWARE"}],
    "devices":
    [{
        "type": MyTuneMonitor.__module__,    
        "class": "MyTuneMonitor",
        "name": "MY_TUNE_MONITOR",
        "control_modes": ["live"],
        "device_h":"TUNEZR:rdH",
        "device_v":"TUNEZR:rdV",
        "frev":1.25e6
    }]
}

# Launch event loop
evloop = asyncio.new_event_loop()
asyncio.set_event_loop(evloop)

# Main code
sr = Accelerator.from_dict(acc_config)
tm1 = sr.live.get_betatron_tune_monitor("MY_TUNE_MONITOR")
print(tm1.tune.get())
print(tm1.frequency.get())
```

```
[0.8474559579144817, 0.7262510844251143]
[1059.3199473931022, 907.8138555313928]
```


JeanLucPons:
It is also possible to attach it and pass a handle to the ring and let the class compute the tune for either a `Simulator` or a `ControlSystem`. Decision here are crucial for the rest of the PyAML objects ?
What do you think ?

TeresiaOlsson:
I can only speak for HZB but we would not use our ophyd-async devices like this. The only thing we need is a configuration layer which is so generic that if we put the class and the required attributes it can build the corresponding python object for each control mode and in that way make it available for the user to use.

We also always measure the tune in all three planes so we need a TuneMonitor which includes the longitudinal plane like we currently have in MML. During the specification discussions it was said that we could have a lab specific device for that. If it's possible in pyAML to build up devices from other devices and have the subdevices automatically be read like in ophyd-async we can could make use of the BetatronTuneMonitor as part of the device, but if not we will just make a separate TuneMonitor.

For the simulator, we will use the soft signal backend in ophyd-async to connect to it.

JeanLucPons:
Here `MyTuneMonitor` is a free class with a free config, you can do what you want and MyTuneMonitor can also ovveride StandardReadable. For 3D tune why not this is just one dimension to add.
You have however to fit pyAML protocols otherwise no way to add them in PyAML.

JeanLucPons:
I can propose something like:

```
config = {
    "class": MyTuneMonitor.__module__+".MyTuneMonitor",
    "name": "MY_TUNE_MONITOR",
    "mode": "live"
    "device_h":"srdiag/beam-tune/main/Qh",
    "device_v":"srdiag/beam-tune/main/Qv",
    "frev":355000.0
}
```

Where i added a mode. Then the object will be added only in the specified control mode and the `ControlSystem` will be passed to allow initialisation with the prefix.

TeresiaOlsson:
Maybe mode then needs to have the possibility to be a list? Otherwise if you want the device for both live and virtual accelerator you need to put the same information twice.

JeanLucPons:
The presence of the `control_modes` property define the `Element` as an `UnboundElement` which will be instantiated when the `ElementHolder` (here MyControlSystem) is filled.

JeanLucPons:
I updated the code sample according to @TeresiaOlsson [example](https://github.com/TeresiaOlsson/ophyd-async-devices/tree/main).
Here `MyTuneMonitor` is both a pyaml Element and an OphydDevice (StandardReadable)
Let me know if it is ok for you.

kparasch (approved):
I suggest we add the example in the PR description as a file in pyaml/examples/ophyd_devices (or with any other folder name).

JeanLucPons:
> I suggest we add the example in the PR description as a file in pyaml/examples/ophyd_devices (or with any other folder name).

I do that

PR #238, opened 2026-04-21
Added Config Dict

JeanLucPons:
This PR allow to have a free dict in a configartion file.

### DIFF
--- a/pyaml/configuration/cfg_dict.py
+++ b/pyaml/configuration/cfg_dict.py
@@ -1,6 +1,26 @@
 import json
 
+from pydantic import BaseModel, ConfigDict
+
 from .manager import ConfigurationManager
+
+# Define the main class name for this module
+PYAMLCLASS = "CfgDict"
+
+
+class ConfigModel(BaseModel):
+    """
+    Configuration model for random dict
+
+    Parameters
+    ----------
+    cfg_dict : dict
+        The dict
+    """
+
+    model_config = ConfigDict(arbitrary_types_allowed=True, extra="forbid")
+
+    cfg_dict: dict
 
 
 class CfgDict(object):
@@ -8,7 +28,7 @@ class CfgDict(object):
     Class allowing to have a dict in a configuration of an object
     """
 
-    def __init__(self, config: dict):
+    def __init__(self, cfg: ConfigModel):
         """
         Construct a ConfigDict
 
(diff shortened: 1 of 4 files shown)


gupichon:
The code is definitely easier this way, but the configuration file requires the user to set the type to pyaml.configuration.cfg_dict, which isn't bound to any pyaml element. It makes the configuration look a bit strange.

I'm not sure a user without any knowledge of the underlying constraints will find it practical, as this type is purely a technical key.

JeanLucPons:
The object is well linked to `pyaml.configuration.cfg_dict.CfgDict` internal pyaml object.
Would you prefer ton construct the dict using the `build_object` function ?
The idea is to be able to write:
```yaml
type: pyaml.configuration.cfg_dict
field1: "field 1 value"
field2: {"param1": "Param1 value", "param2": 12345.0}                },
```

gupichon:
Actually, I find the current approach counter-intuitive. In the existing version, you can define a model using a list in pyaml:

```python
class ConfigModel(BaseModel):
    multipoles: list[str]
    curves: list[Curve]
```

This corresponds to the following configuration:

```yaml
type: pyaml.my.model
  multipoles: ["this", "is", "a", "list"]
  curves:
    - curve1.yaml
    - curve2.yaml
```

I believe we should apply the exact same principle to dictionaries:

```python
class ConfigModel(BaseModel):
    data: dict[str]
```

```yaml
type: pyaml.my.model
  data:
    my_data1: data
    my_data2: data
```

This would make the API consistent across all collection types.

JeanLucPons:
In you example the curves yaml files are expanded and curve are constructed first to construct the list of Curve
A dict is a random structure with a random depth, to me, your proposal does not wok.
You have to specify a type that will stop the recursion.

I did a mods to contruct in a classic way but it change to:
```yaml
type: pyaml.configuration.cfg_dict
cfg_dict:
  field1: "field 1 value"
  field2: {"param1": "Param1 value", "param2": 12345.0}
```

gupichon (approved):

gupichon:
I understand it's complicated. However, it feels counterintuitive to use a keyword for something that seems trivial. That said, your solution is the best compromise for now.

JeanLucPons:
I agree.
I'm searching a solution but fightting with pydantic field is a nigthmare.
But it is progressing

JeanLucPons:
Bad news pydantic use TypedDict for dict and it is not testable with isinstance() :(
A work around could be to check `<class 'dict'>`

PR #240, opened 2026-04-21
Add shared dummy Tango attribute state

### LINKED ISSUE #239: Feature: shared Tango attribute state in dummy control system
The test dummy control system currently creates independent `DeviceAccess` objects for Tango attributes. This makes it difficult to prepare a test context where a Tango attribute has a predefined value before the accelerator configuration is loaded.

For tests, dummy Tango attributes should behave like shared objects identified by their full attribute name. If a test initializes a value for a given attribute, every `DeviceAccess` object created later for the same attribute should read and update that same shared value.

This is especially useful for BPM tests based on array-valued Tango attributes. A BPM position may be backed by one Tango attribute returning both horizontal and vertical values. Today, there is a test-specific trick in the dummy control system: `attach_array()` marks the attached attribute as an array, and `Attribute.set_array(True)` initializes its internal cache to `[0.0, 1.0]`. This implicit initialization should be removed and replaced by explicit test setup.

This feature should be implemented from `main` and should not depend on catalog-related changes.

**Proposed solution**
Add a shared registry in the dummy Tango package, keyed by full Tango attribute name.

Each registry entry should store:

- current value
- readback value, if needed
- unit
- range
- read-only/read-write behavior, if needed

`Attribute` instances should no longer own their value directly. Instead, they should delegate `get()`, `set()`, `readback()`, `unit()`, and `get_range()` to the shared registry entry associated with their full attribute name.

Expose test helper functions such as:

```python
set_attribute(name: str, value, unit: str = "", range=None)
get_attribute(name: str)
clear_attributes()
```

`clear_attributes()` should be used between tests to avoid state leakage. Pytest already provides mechanisms for this: a fixture can be added in `conftest.py` to automatically clear the dummy Tango registry before or after each test, so individual tests do not have to call `clear_attributes()` manually.

The current alternative is to keep initializing array attributes implicitly from `attach_array()` through `Attribute.set_array(True)`. This is not ideal because the test data is hidden in the dummy implementation instead of being explicit in the test setup.

Another alternative would be to initialize each `Attribute` instance independently. This does not model Tango-like behavior correctly when several access objects point to the same attribute, and it does not allow a value prepared before `Accelerator.load()` to be shared by all objects created afterward.

**Example**
Example test setup:

```python
dummy_tango.set_attribute("srdiag/bpm/c01-04/Position", [0.0, 1.0], unit="mm")
```

Then any dummy `Attribute` created later for `srdiag/bpm/c01-04/Position` should read the initialized array value.

This would allow the indexed BPM control-system test to prepare the expected array explicitly instead of relying on the current implicit `[0.0, 1.0]` initialization in `Attribute.set_array(True)`.

**Additional context**
Expected behavior:

- A test can set the initial value of `srdiag/bpm/c01-04/Position` before loading the accelerator.
- Any dummy `DeviceAccess` created afterward for `srdiag/bpm/c01-04/Position` reads that initialized value.
- If one access object writes a new value, all other access objects for the same attribute see the updated value.
- Read-only attributes can still be initialized by the test setup.
- Attributes can be initialized either as scalars or arrays.

Acceptance criteria:

- Two dummy `Attribute` instances created with the same Tango attribute name share the same value.
- A value initialized before `Accelerator.load()` is visible from the loaded accelerator.
- Array-valued attributes are supported.
- Read-only attributes can be initialized by test setup.
- Tests can reset the dummy Tango registry between runs, preferably through a pytest fixture in `conftest.py`.

**Checklist**
- [X] I've assigned this issue to a project

gupichon:
## Description

Add shared state for dummy Tango attributes used in tests. This allows tests to initialize an attribute value before loading a configuration, and all later dummy `DeviceAccess` objects for the same Tango attribute share that value.

## Related Issue

- Fixes #239

- [x] new feature: dummy Tango attributes now use a shared registry keyed by attribute name.

- [x] `tango.pyaml.attribute` delegates value, unit, range and readback storage to a shared registry.
- [x] `attach_array()` no longer relies on implicit `[0.0, 1.0]` initialization; tests initialize array attributes explicitly.
- [x] pytest clears the dummy Tango registry between tests.

## Testing

The following tests were added or updated:
- [x] shared dummy Tango attribute state
- [x] read-only attribute initialization
- [x] indexed BPM test with explicit array initialization


JeanLucPons:
On the code you write (not only on this PR) there is absolutely no comment. May be your code is clear for you but I often pain to read it. Please try to put few comments to explain a bit what the code do ! Few lines can help a lot other to understand and to maintain.
Last but not least there are lots of other priority task to do.
Thanks for understanding.

gupichon:
I'll add comments. I actually need this to simplify the BPM evolution and tests for other developments. For example, the `attach_array` method will be removed since it’s no longer necessary (it simply calls `attach` on the backends).

Moreover, it will simplify BPM testing as you’ll be able to set them to test values in live mode. This will allow for tests on the live system that were previously impossible.

gupichon:
By the way, I added a test for dummy-cs. It’s not mandatory, but it will help since it’s something we actually have to maintain..

JeanLucPons:
> I'll add comments. I actually need this to simplify the BPM evolution and tests for other developments. For example, the `attach_array` method will be removed since it’s no longer necessary (it simply calls `attach` on the backends).
> 
Not on the pyaml_cs_oa side, but this should change if the Ophyd Asynch fix is deployed.

JeanLucPons:
```python
    def attach(self, devs: list[OASignal]) -> list[OASignal]:
        return self._attach(devs,False) # is_array=False

    def attach_array(self, devs: list[OASignal]) -> list[OASignal]:
        return self._attach(devs,True) # is_array=True
```

gupichon:
> Not on the pyaml_cs_oa side, but this should change if the Ophyd Asynch fix is deployed.

I forgot about this point, you're right. Anyway, it will help a lot with testing, as you can set the live part with this development.

JeanLucPons:
Is it ready to merge ?

gupichon:
> Is it ready to merge ?

Yes, all tests are passing, and I have added a test specifically for the dummy because, ultimately, we depend on it for delivery.

JeanLucPons (approved):

PR #264, opened 2026-05-20
Add a schema registry

TeresiaOlsson:
This PR adds a schema registry that can be used for validation and generating json schemas for dynamic nested pydantic models.

**Motivation:**

- Better separation of concerns. Validation of the configuration is separated into different classes than the ones responsible for storing and using the configuration, resulting in a separate validation layer.

- It is possible to make validation option. The entire validation layer can be skipped if the user wished. For example, if they know that the input data has already been validated once and not changed.

- More lose dependency of pydantic. Pydantic is only used at the edge of the core and not everywhere in line with what has been discussed for how to manage the dependency of pint.

- Solving two problems with pydantic:
  1. It is no good at handling nested models where the available types are only known at runtime.
  2. It is no good at generating json schemas for arbitrary types and including all available subclasses of a model.

**Features:**

- A schema registry which maps a class path to the schema that should be used for validation of the input data to the class.

- A decorator to automatically register schemas in the schema registry.

- A schema validator which validates nested dynamic models with the use of the registry.

- A custom json schema generator which generates json schemas including all available subclasses with the use of the registry.

**Major changes:**

- All ConfigModels have been separate from the domain classes and turned into schema classes which only has the purpose to define the schema used for validation. This is a major refactoring since the ConfigModels are spread out in the codebase. A temporary legacy handler had to be implemented to handle the packages for the bindings. All domain classes also had to be changed since they now need to explicitly declare their attributes.

- A baseclass `ConfigurationSchema` has been added that all schemas must inherit from. This is to ensure that all schemas registered in the registry has the minimum required fields and same behaviour.

- The yaml file format has been changed to instead of defining the type by the module it is defined by the class. This is to be compatible with external packages where a certain module structure can't be enforced and to allow several similar classes to be grouped in the same module. A temporary function to translate between yaml file formats have been added.

**Example usage:**
Examples of the usage is available in [https://github.com/python-accelerator-middle-layer/pyaml/tree/schema-registry/examples/validation](https://github.com/python-accelerator-middle-layer/pyaml/tree/schema-registry/examples/validation).

The json schema can be visualized and tested with [MetaConfigurator](https://www.metaconfigurator.org/) to easier understand what it includes. You can import the schema directly from here: [https://github.com/python-accelerator-middle-layer/pyaml/blob/schema-registry/pyaml/validation/schema.json](https://github.com/python-accelerator-middle-layer/pyaml/blob/schema-registry/pyaml/validation/schema.json)


JeanLucPons (dismissed):

JeanLucPons:
I do not understand why some external modules appears in committed json files ?
These files are generated ?

TeresiaOlsson:
I will explain it better on the meeting tomorrow. It's a temporary quite ugly fix I added to be able to register the schemas for the binding packages and convert the format of the yaml files. There is also one related ugly thing in the pyproject.toml at the moment. I didn't manage to figure out a less ugly way to handle it.

My plan is that all of that should go away when people have converted the yaml file format and I have updated the packages for the bindings to instead use entry points to find the schemas in there.

JeanLucPons:
OK thanks

TeresiaOlsson:
For the json schema I committed the full schema including the external packages to be able to provide a link which people can directly load in the metaconfigurator for testing without having to generate it themselves.

But I'm not sure if there should be a "standard" schema available in the repository in the future or not. I kind of like the idea of having a pregenerated schema including the most common external packages for new users to get started without having to generate the schema themselves. Then you only need to generate your own schema if you have facility specific packages. But if there should be such a basic, standard schema maybe it should be in it's own repository. Then we in that repository could also have different files for different parts of the schema if we want.

JeanLucPons:
OK so it would be nice to provide a way to generate this schema with the modules you want to use.
Using catalog it should be possible to configure your backend using only simple strings.

TeresiaOlsson:
That I think already works. As long as you have everything you want to include registered in the schema registry it will show up in the json schema. The key to it is to get external packages to automatically register in the schema registry and make that compatible with what the catalog needs. Those parts doesn't automatically work yet and might require some discussion to figure out the best way since I'm still working on understanding the details of the catalog to be able to make it compatible. But at least you can always manually register everything you want to include.

JeanLucPons:
I'm starting to look at your implementation.
If i understand well your strategy, your plan is too remove pydantic validation from the Factory and let the factory expand dictionary to construct objects ?
If I'm right, please think how to handle error line number in validation. (see Factory.handle_validation_error).
(Your validation mechanism should handle `__location__` and `__fieldlocation__` private fields coming from the loader)
Thanks.

JeanLucPons:
For me the purpose of Adapted_ControlSystem is a bit unclear and would it be possible to have one schema per magnet type to avoid those unwanted layers (item1->item1) ?

image

TeresiaOlsson:
> I'm starting to look at your implementation. If i understand well your strategy, your plan is too remove pydantic validation from the Factory and let the factory expand dictionary to construct objects ? If I'm right, please think how to handle error line number in validation. (see Factory.handle_validation_error). (Your validation mechanism should handle `__location__` and `__fieldlocation__` private fields coming from the loader) Thanks.

Yes, my plan is to make the validation a completely separate step done before the factory so the factory is only responsible for building objects. I'm aware of the error line number and also the function to reformat the validation errors from pydantic so they become easier to understand. I want to move that functionality into the SchemaValidator as part of refactoring the factory.

JeanLucPons:
> Yes, my plan is to make the validation a completely separate step done before the factory so the factory is only responsible for building objects. I'm aware of the error line number and also the function to reformat the validation errors from pydantic so they become easier to understand. I want to move that functionality into the SchemaValidator as part of refactoring the factory.

OK. The refurbishment of the factory should ne be a big deal. You only need to use the new class field to construct the object and expand the dictionary. What I dislike here is the duplication of the constructor signature (as discussed long time ago) which can be an error source difficult to debug. But no other way to allow construction by code without having to construct the schema first.

TeresiaOlsson:
> For me the purpose of Adapted_ControlSystem is a bit unclear and would it be possible to have one schema per magnet type to avoid those unwanted layers (item1->item1) ?
> image

The Adapted_ControlSystem is part of the ugly fix to be able to include the ConfigModel from the binding packages despite that they don't follow the new format and inherit from ConfigurationSchema which is a requirement to be added into the SchemaRegistry. It will go away as soon as I refactor the schemas in those packages. They are just called Adapted_ControlSystem at the moment to indicate that it is part of the legacy fix.

For the magnets I also don't like the two layers of item1->item1 in the GUI but I found no way around it. Before there was a separate schema for each magnet type but it just meant I had to choose OctupoleSchema in the first box and still choose Octupole as class since the GUI doesn't prefill the class even if there is only one option. In the end I concluded that these are issues related to the metaconfigurator GUI which can be fixed in the future either by writing our own GUI application or make feature requests to the metaconfigurator. Instead I decided to follow the principle to only add a new schema if there is a field which makes it different. My hope is that eventually that would allow us to have oneOf instead of anyOf in the json schema which also would make the GUI work nicer. But turns out that oneOf is a very strict requirement and it isn't sufficient to just have two schemas with different fields if the fields that make them different are all optional. They need to have mandatory fields which are different and that might be a too strict requirement for us. But I'm still thinking about that.

TeresiaOlsson:
> > Yes, my plan is to make the validation a completely separate step done before the factory so the factory is only responsible for building objects. I'm aware of the error line number and also the function to reformat the validation errors from pydantic so they become easier to understand. I want to move that functionality into the SchemaValidator as part of refactoring the factory.
> 
> OK. The refurbishment of the factory should ne be a big deal. You only need to use the new class field to construct the object and expand the dictionary. What I dislike here is the duplication of the constructor signature (as discussed long time ago) which can be an error source difficult to debug. But no other way to allow construction by code without having to construct the schema first.

I also disliked the duplication of the constructor signature initially. It was annoying to have to write the same thing twice. I tried some ideas to avoid it but found none that I liked and which was easier to use than copy-paste. But after a while I actually started to like it because it made it explicit what the attributes are for building the domain objects.

I found several cases where there were attributes in the ConfigModel which was never used for anything because they were hidden inside the `_cfg` and never exposed as public attributes in the domain class. Having to write all the attributes in the constructor forced me to think about which attributes should be private/public, which need properties because they should be read only or need some extra step when setting etc. For new users/developers I imagine that this is beneficial because it forces them to think about what the public interface of the class should be. Then when they are done figuring that out they can add the schema or register an already existing schema for their class.

I also liked how it forced me to think about the scope of the class. In a pydantic basemodel it is very easy to add 10 attributes but when you need to add all of them again to `__init__` you start to realise that maybe they are too many. I think many attributes can be an indication of a class having a too large scope and that it should be separated into smaller classes to get a better separation of concerns.

JeanLucPons:
My main problem is that this duplication can be an error source if you forget one parameter or enter them in a wrong order you may face tricky issue to solve. Duplication is generally not a good idea.
Today `_cfg` is supposed to be private which means that you are supposed to access it only from pyaml core. But _cfg should not be accessed by the user. I would prefer for dynamic configuration change to add some public methods. Anyway, in your schema, additional methods for changing the config will also be required.
i.e.

Today:

```python
# Set up the  chromaticy monitor (override config settings)
CM = SR.get_chromaticity_monitor("CHROMATICITY_MONITOR")
CM._cfg.n_step = 3      # 3 point for chroma fit
CM._cfg.n_avg_meas = 1  # No averaging
```

should be:

```python
# Set up the  chromaticy monitor (override config settings)
CM = SR.get_chromaticity_monitor("CHROMATICITY_MONITOR")
CM.set_n_step(3)      # 3 point for chroma fit
CM.set_n_avg_meas(1)  # No averaging
```

TeresiaOlsson:
> My main problem is that this duplication can be an error source if you forget one parameter or enter them in a wrong order you may face tricky issue to solve. Duplication is generally not a good idea. Today `_cfg` is supposed to be private which means that you are supposed to access it only from pyaml core. But _cfg should not be accessed by the user. I would prefer for dynamic configuration change to add some public methods. Anyway, in your schema, additional methods for changing the config will also be required. i.e.
> 
> Today:
> 
> ```python
> # Set up the  chromaticy monitor (override config settings)
> CM = SR.get_chromaticity_monitor("CHROMATICITY_MONITOR")
> CM._cfg.n_step = 3      # 3 point for chroma fit
> CM._cfg.n_avg_meas = 1  # No averaging
> ```
> 
> should be:
> 
> ```python
> # Set up the  chromaticy monitor (override config settings)
> CM = SR.get_chromaticity_monitor("CHROMATICITY_MONITOR")
> CM.set_n_step(3)      # 3 point for chroma fit
> CM.set_n_avg_meas(1)  # No averaging
> ```

I'm not sure I understand. In the schema there should be no method to change the configuration? The purpose of the schema classes is purely to define which fields are required for validation. You would then use the `SchemaValidator` to validate the data but the basemodel that is returned will not be used for anything else than extracting a dictionary with the validated data to be input to other classes. It is not meant to be stored. The parts which includes how to store and dynamically change the configuration is outside of the scope of this PR.

However, if people like the idea, it is possible to later make an implementation where dynamic changes of the configuration also calls the `SchemaValidator` to validate the data again before it is changed. I have started to think about that but the PR became too large so I decided to split it and make one first which is only focused on the validation step.

JeanLucPons:
My remarks was rather link on the the consequence of your validation model (not on the schema itself) and the fact that fields have to be duplicated at the object level which means that you have in fact 3 duplications:
- model schema
- object constructor signature
- object fields to access the configuration

In an ideal work, to follow pyaml coding style, i would like to be able to write:

```python
# Set up the  chromaticy monitor (override config settings)
CM = SR.get_chromaticity_monitor("CHROMATICITY_MONITOR")
CM.n_step.set(3)      # 3 point for chroma fit
CM.n_avg_meas.set(1)  # No averaging
```

and having a mechanism (using a decorator or dynamic code generation) to map automatically schema fields to object getter(s)/setter(s) with an optional callback which allow the object to be informed of field updates. It will require at the schema level to be able to select if a field is R or RW.

TeresiaOlsson:
> My remarks was rather link on the the consequence of your validation model (not on the schema itself) and the fact that fields have to be duplicated at the object level which means that you have in fact 3 duplications:
> 
>     * model schema
> 
>     * object constructor signature
> 
>     * object fields to access the configuration
> 
> 
> In an ideal work, to follow pyaml coding style, i would like to be able to write:
> 
> ```python
> # Set up the  chromaticy monitor (override config settings)
> CM = SR.get_chromaticity_monitor("CHROMATICITY_MONITOR")
> CM.n_step.set(3)      # 3 point for chroma fit
> CM.n_avg_meas.set(1)  # No averaging
> ```
> 
> and having a mechanism (using a decorator or dynamic code generation) to map automatically schema fields to object getter(s)/setter(s) with an optional callback which allow the object to be informed of field updates. It will require at the schema level to be able to select if a field is R or RW.

If you want that I think it should be implemented in a way to make the object constructor signature being the source of truth. There is a way in pydantic to do that using `TypeAdapter` but when I tried it I didn't really like it. In the end my conclusion was that the duplication was the more robust way to implement it. Especially after I concluded that there are many classes which require the exact same input so there is no need to have separate schemas.

JeanLucPons:
OK. For me it is a bit heavy to have all theses duplications and also loose the repr from pydantic.

For instance this will not work:
```python
    arrays: list[ArraySchema] = Field(default_factory=list, repr=False)
    devices: list[ElementSchema] = Field(default_factory=list, repr=False)
```

Anyway, this refurbishment is heavy and we need to stabilize the implementation ASAP.

TeresiaOlsson:
The only option I have found to avoid the duplication which I slightly believe in is discussed here: https://stackoverflow.com/questions/65888153/creating-a-pydantic-model-dynamically-from-a-python-dataclass

That would define dataclasses in the business logic which would be dynamically translated into Pydantic. I think the modern way to do that would be to use [TypeAdapter]( https://pydantic.dev/docs/validation/latest/api/pydantic/type_adapter/).

But in the end my conclusion after considering different options is the same as someone else also has in the comments. It's better to just write it twice because over time the internal model and the validation schema might start to differ.

I think we already have started to see that. Validation and generating the json schema doesn't work as well when arbitrary types are allowed, but it wouldn't make sense to restrict the domain classes to not take custom classes as input. For the majority of the schema classes I have introduced there is duplication of the attribute names but not of their type for exactly this reason. The validation schema and the internal model works better when they are different because they have different purposes and therefore different requirements.

JeanLucPons:
I didn't find time yet at the moment to investigate on this issue.

The duplication is not acceptable for me as there are in fact 4 duplications (not 3).

- model
- constructor signature
- object field for repr and edition
- attachment process

I was rather thinking to introspect the pydantic schema and when an arbitrary type is found, resolve the corresponding model in order to be able to construct a full JSON schema. I don't know yet if pydantic model introspection is easily feasible.

TeresiaOlsson:
As far as I have understood the only way to get it to work with arbitrary types is to implement one of the options here: https://pydantic.dev/docs/validation/latest/concepts/json_schema/#customizing-json-schema. I think `__get_pydantic_core_schema__` is also the one that needs to be implemented to be able to validate arbitrary types. Without it, the validation is limited to just checking that the type is correct and it doesn't validate the data inside the class. It can probably be implemented in some generic way if all the ConfigModels/schemas are forced to inherit from a common base rather than BaseModel.

The result in the end would however still be that the business logic of the core is dependent on pydantic. Part of my suggestion is also to make it a dependency just on the edge of the core in a validation layer which can be replaced or skipped completely if there are problems in the future. So the question for me is how strong we want the dependency of pydantic to be? If we decide to keep the strong dependency, one option is also not to have a separate ConfigModel/schema at all but just make all devices inherit from BaseModel and they will get the functionality automatically. I think that's what they do in CATAP but we can take a look at the details of their implementation here: https://github.com/astec-stfc/acronicta-catap

TeresiaOlsson:
@JeanLucPons If you think it is a good idea, I can try to make separate branches from this PR to explore different options how it could be implemented without duplication. I think most of the PR would be unaffected. It only relies on the classes that are registered in the SchemaRegistry to have certain functionality.

JeanLucPons:
> So the question for me is how strong we want the dependency of pydantic to be?

The less possible. For the core, It is very important to minimize dependencies to external lib as much as possible.
Otherwise it will be more difficult and slower to maintain the core code.

TeresiaOlsson:
> > So the question for me is how strong we want the dependency of pydantic to be?
> 
> The less possible. For the core, It is very important to minimize dependencies to external lib as much as possible. Otherwise it will be more difficult and slower to maintain the core code.

I agree. It would be great if we somehow could have external dependencies mostly on the edge of the core. But I'm not sure how to do that for pydantic without getting some duplication of attributes. But maybe there is a way with just some more thinking and trying different options.

JeanLucPons:
I'm investigating on your PR and how to hanlde arbitratry type.  __get_pydantic_core_schema__ looks promising.
https://pydantic.dev/docs/validation/latest/concepts/json_schema/#implementing-__get_pydantic_core_schema__

JeanLucPons:
I managed to get something working.
I don't know if it produce a usable schema.
But at least there are definitely something that we can do with arbitrary type.

```python
from dataclasses import dataclass
from typing import Any
from pydantic_core import core_schema
from pydantic import BaseModel, GetCoreSchemaHandler, GetJsonSchemaHandler, ConfigDict
from pydantic.json_schema import JsonSchemaValue
from pydantic_core import core_schema as cs

class MyClass1ConfigModel(BaseModel):
    param1:str
    param2:int

class MyClass1:

    def __init__(self,cfg:MyClass1ConfigModel):
        self._cfg = cfg
    
    @classmethod
    def __get_pydantic_json_schema__(
        cls, core_schema: cs.CoreSchema, handler: GetJsonSchemaHandler
    ) -> JsonSchemaValue:
        # Map the schema of the config object
        sch = MyClass1ConfigModel.__get_pydantic_core_schema__(MyClass1ConfigModel,handler)
        print(sch)
        return sch
        
class MyClass2ConfigModel(BaseModel):
 
    model_config = ConfigDict(arbitrary_types_allowed=True, extra="forbid")

    param:str
    extern:MyClass1

class MyClass2:

    def __init__(self,cfg:MyClass2ConfigModel):
        self._cfg = cfg


mycfg = MyClass1ConfigModel(param1="Param1Value",param2=10)
mc1 = MyClass1(mycfg)
mycfg2 = MyClass2ConfigModel(param="Param2Value",extern=mc1)
mc2 = MyClass2(mycfg2)
print(mc2)

print(MyClass2ConfigModel.model_json_schema())
```

JeanLucPons:
I just realize that this way of doing is deprecated :(

JeanLucPons:
I'm a bit disappointed that pydantic deprecates this intuitive and well documented method and that they do not provide clear  equivalent code. handler(source) is far from being clear !

```
/home/esrf/pons/pyaml/tests/jlp_test.py:24: PydanticDeprecatedSince211: The `__get_pydantic_core_schema__` method of
the `BaseModel` class is deprecated. If you are calling `super().__get_pydantic_core_schema__` when overriding the method 
on a Pydantic model, consider using `handler(source)` instead. However, note that overriding this method on models can 
lead to unexpected side effects. Deprecated in Pydantic V2.11 to be removed in V3.0.
```

gupichon:
I took a look at the PR and, in my opinion, the duplication of parameters is annoying, but it's the lesser of two evils (or the best of the worst options). Also, configuration validation can be done within the configuration classes using specific methods.

Moreover, having a strong separation between the functional classes and the configuration classes by not passing them as parameters is a very good way to avoid depending too much on Pydantic. It will ease the development of other loaders (database, facility-specific configuration files, ...) since it will be much easier to build an accelerator programmatically.

gupichon:
With this PR, I can develop a loader for SOLEIL that directly builds the accelerator from the SOLEIL lattice description file, instead of having to create a PyAML configuration file. That said, I would still like to keep the PyAML configuration file for small configurations.

JeanLucPons:
We have a working solution without duplication and that handle repr in a simple and efficient manner.
The only missing added value is the json schema but I would do in a different manner because writing an expert tools for pyaml configuration based on json schema will be a nightmare.

There are several problems with the actual proposal:
- 4 duplications of the parameters (Model, constructor, fields for repr, attachment). This is too much.
- Strong dependance to pydantic with an overriding of __get_union_of_schema__ and __model_schema__ that can end in same problem than overriding __get_pydantic_core_schema__
- Apparently no way to handle same schema with same signature and a different name that will be a problem with pyaml_cs_oa

However lots of good other points.... I agree... 
Personally, I can already build a full configuration file (or dict) from our lattice files. At ESRF we even already have device name in our lattice file but this is not a standard.

gupichon:
As the PyAML configuration is highly extensible, adopting a JSON schema should remain optional or contextual. Each facility will have its own schema, as I understand it, so I don't see how we can provide a standard schema for the PyAML configuration, though I might be missing something.

Regarding the duplication of parameters: this is a common pattern when dealing with DTOs, and it serves a clear purpose for the separation of concerns. I'm used to it, so it doesn't seem like a major issue here.

Finally, I might be missing the specific issue with `__repr__`, but we could achieve the exact same behavior as today by leveraging the decorator instead of relying on a stored configuration object, which might become outdated or insufficient in certain scenarios.

gupichon:
Honestly, I'm very interested in the possibility of instantiating an accelerator directly, without the overhead of creating a configuration file or defining config model objects. Forcing these steps creates a steep learning curve, which is something non-developers especially find very difficult to navigate in my experience.

TeresiaOlsson:
I was also looking into a solution with `__get_pydantic_core_schema__` but didn't realise that it was deprecated. Information about that is also not clear in the documentation... Otherwise it also felt to me like the best option.

That would however only solve the problem with the JSON schema and being able to validate arbitrary types. The other problem is that the ConfigModel doesn't contain all the required fields for validating the configuration and generating the JSON schema. For that there has to also be a mandatory field `type/class`. So then we need to add that field to every ConfigModel before registering it the schema registry. That is currently what I do in the legacy module to be able to register the ConfigModels from the bindings packages and I found it to be somewhat ugly.

For me that's why I don't really think about it as duplication. Most of the attributes are the same, but not all of them. At the moment I think we have two concerns in the same class, validating the configuration and defining the attributes for the domain object and those two don't have entirely the same requirements.

For the `___repr___` I have been working on a solution to replace it with a function that prints the attributes based on the public attributes and properties of the class. So far it has worked quite well but I haven't fully finished testing it yet.

JeanLucPons:
You may need to duplicate signature in some scenario but here it is not mandatory. It has already been thoroughly discussed and finally Yoshi found the clean solution.
Last but not least this affects some basis already discussed in past and will lead to a big refurbishment and potential problems.
Anyway,  I will **never** accept 4 duplications. This is a too strong error source !

TeresiaOlsson:
> * Strong dependance to pydantic with an overriding of **get_union_of_schema** and **model_schema** that can end in same problem than overriding **get_pydantic_core_schema**

For me it's not a strong dependency because it's just in the validation layer. The validation layer however is completely dependent on Pydantic but that is just because we decided to use Pydantic for validation. One of the things I wanted to achieve with the PR is to make it easy in the future to replace Pydantic if it turns out to be necessary. With this PR the validation is done in a separate step so it is possible to write a new validation layer based on something else and replace the Pydantic based one without affecting the business logic parts of the code.

JeanLucPons:
> That would however only solve the problem with the JSON schema and being able to validate arbitrary types. The other problem is that the ConfigModel doesn't contain all the required fields for validating the configuration and generating the JSON schema. For that there has to also be a mandatory field `type/class`. So then we need to add that field to every ConfigModel before registering it the schema registry. That is currently what I do in the legacy module to be able to register the ConfigModels from the bindings packages and I found it to be somewhat ugly.

Yes I agree. I'm sure it is possible to keep a ConfigModelSomething structure as before and register it. I'm ok to have different names to be able to work by class and not by module. I have nothing against this point and your SchemaRegistry. I'm just a bit afraid of deprecation of pydantic stuff.

> For the `___repr___` I have been working on a solution to replace it with a function that prints the attributes based on the public attributes and properties of the class. So far it has worked quite well but I haven't fully finished testing it yet.

I don't clearly see how you want to handle this. Let's see.

JeanLucPons:
> For me it's not a strong dependency because it's just in the validation layer. The validation layer however is completely dependent on Pydantic but that is just because we decided to use Pydantic for validation. One of the things I wanted to achieve with the PR is to make it easy in the future to replace Pydantic if it turns out to be necessary. With this PR the validation is done in a separate step so it is possible to write a new validation layer based on something else and replace the Pydantic based one without affecting the business logic parts of the code.

At the beginning I proposed a simple decorator that handle validation but it has been strongly refused and i have been forced to accept pydantic. There was no duplication and the source of truth was the constructor of the object.
I can accept the above and forget pydantic with pleasure :)

gupichon:
I completely agree with @TeresiaOlsson. Taken to its logical conclusion, the configuration layer (and Pydantic) could even live in an entirely separate project, completely decoupled from the PyAML core. While perfectly possible, that might be overkill for our current scope.

Moreover, parameter duplication is minimal here: you don't need to maintain separate fields for `__repr__`, and the attachment process shouldn't require duplication either (it as to be checked). 

If you look at how other major projects leverage Pydantic (LangChain and LangGraph, for example, which I'm currently working with on another project), they adopt this exact pattern.

TeresiaOlsson:
> Honestly, I'm very interested in the possibility of instantiating an accelerator directly, without the overhead of creating a configuration file or defining config model objects. Forcing these steps creates a steep learning curve, which is something non-developers especially find very difficult to navigate in my experience.

Me too. I also even want to be able to create a single device directly without having to load any configuration or the rest of pyAML. I think this is the process most users will have when developing a new device that they want to add to pyAML:

1. First they develop the device on it's own. They load the required modules from pyAML and start writing the class for the new device. They put the configuration data directly into the python code and build an object of the class. Many of them might use jupyter notebooks and play with the device in an iterative way until the logic works like they want and they can set/put to the machine or twin.

2. They now want to add the new device to pyAML and part of the Accelerator. At this point they will realise that they want to be able to configure it using the yaml file and have the input data validated. Then they will write the schema for it.

That is based on observing what our users do when doing development using pyepics and ophyd-async. They are all physicists and in my experience they think first about the part that "is doing something" because they are eager to test on the machine. Only later when that works they start to think about things that are important for the code quality like data models and validation.

gupichon:
> Me too. I also even want to be able to create a single device directly without having to load any configuration or the rest of pyAML. I think this is the process most users will have when developing a new device that they want to add to pyAML:
> 
>     1. First they develop the device on it's own. They load the required modules from pyAML and start writing the class for the new device. They put the configuration data directly into the python code and build an object of the class. Many of them might use jupyter notebooks and play with the device in an iterative way until the logic works like they want and they can set/put to the machine or twin.
> 
>     2. They now want to add the new device to pyAML and part of the Accelerator. At this point they will realise that they want to be able to configure it using the yaml file and have the input data validated. Then they will write the schema for it.
> 
> 
> That is based on observing what our users do when doing development using pyepics and ophyd-async. They are all physicists and in my experience they think first about the part that "is doing something" because they are eager to test on the machine. Only later when that works they start to think about things that are important for the code quality like data models and validation.

I have observed the exact same workflow and received identical feedback from Jupyter users at SOLEIL.

TeresiaOlsson:
> > For me it's not a strong dependency because it's just in the validation layer. The validation layer however is completely dependent on Pydantic but that is just because we decided to use Pydantic for validation. One of the things I wanted to achieve with the PR is to make it easy in the future to replace Pydantic if it turns out to be necessary. With this PR the validation is done in a separate step so it is possible to write a new validation layer based on something else and replace the Pydantic based one without affecting the business logic parts of the code.
> 
> At the beginning I proposed a simple decorator that handle validation but it has been strongly refused and i have been forced to accept pydantic. There was no duplication and the source of truth was the constructor of the object. I can accept the above and forget pydantic with pleasure :)

I think there is a lot of power with using pydantic. It makes it easy to add custom validation and generate a JSON schema without having to write that code from scratch yourself. However, it seems to be a common problem with pydantic that is starts to go everywhere in the code and I have found several other projects where they also are discussing the same issue as we are.

I would prefer the constructor of the object to be the source of truth and that the pydantic basemodels only is something extra that is used in a separate validation layer that you can skip if there is an issue. Do you still have an example of the solution you had before? I think I never saw it. If you in that solution had the constructor as source of truth and already had solved the problems with the `__repr__` without needing pydantic one option would be to go back to that + combined with this PR where pydantic is separate and can be optional. Then only the ones who want to use pydantic needs to deal with the duplication.

gupichon:
I also think it shouldn't be possible to create an element outside an element holder context. It’s a major source of confusion. Currently, during file loading, element classes handle configuration validation before being cloned to become the final operational objects. This lifecycle was really confusing when I first encountered it. 

Elements should only be instantiated within the scope of element holders, not leveraged for validation. Shifting this responsibility would also cleanly eliminate an unnecessary duplication step.

JeanLucPons:
Again.
At the beginning I proposed a simple decorator that handle validation but it has been strongly refused and I have been forced to accept pydantic which is bad with nested objects. It was not at all my choice and yes my point is also to get something working and not to refurbish thing each time otherwise we will progress too slowly.
You have to understand that changing fundamental decision has strong impact and slow down the development even if the past decision are not so good. Do what you want, either you simplify thing and you start to work on important stuff either I will have to step down.

gupichon:
> Again. At the beginning I proposed a simple decorator that handle validation but it has been strongly refused and I have been forced to accept pydantic which is bad with nested objects. It was not at all my choice and yes my point is also to get something working and not to refurbish thing each time otherwise we will progress too slowly. You have to understand that changing fundamental decision has strong impact and slow down the development even if the past decision are not so good. Do what you want, either you simplify thing and you start to work on important stuff either I will have to step down.

I completly understand your point, but the steering committee has actually stated that simplification is the current priority. Am I wrong, @TeresiaOlsson?

JeanLucPons:
> I completly understand your point, but the steering committee has actually stated that simplification is the current priority. Am I wrong, @TeresiaOlsson?

Simplification of the configuration.

TeresiaOlsson:
> aken to its logical conclusion, the configuration layer (and Pydantic) could even live in an entirely separate project, completely decoupled from the PyAML core. While perfectly possible, that might be overkill for our current scope.

I like the suggestion for it to live in a separate project. My feeling is that we start by having everything inside the pyaml-core but over time we should move towards having things in many separate repositories and build up an ecosystem of packages and learn how to manage that type of environment. It will be a learning curve because we need to learn how to manage dependencies etc but it would give us the type of environment where labs can pick and choose the parts they like and add their own development without having to care about fulfilling the requirements of everyone else.

JeanLucPons:
> . Do you still have an example of the solution you had before? 
It was at the very beginning it was a @validate decorator on the constructor.

JeanLucPons:
I would really prefer that you work on more important stuff rather than fighting with pydantic:

- pyaml device abstraction
- diagnostic devices such as emittance, beamloss, TbT, ...
- disabled device
- tuning/measurement tools (learn pySC and implement BBA, LOCO, ...)
- Expert application for config (something similar to Jive)
- Unit conversion
- Documentation

TeresiaOlsson:
> > I completly understand your point, but the steering committee has actually stated that simplification is the current priority. Am I wrong, @TeresiaOlsson?
> 
> Simplification of the configuration.

The roadmap includes both the configuration and the hardware abstraction layer. Configuration is the main priority until the summer and then the hardware abstraction layer until the end of the year. The full roadmap is here: https://github.com/python-accelerator-middle-layer/governance/blob/main/roadmaps/roadmap_2026-2027.pdf

At least from my point of view (but we can discuss at the next steering committee since I can't speak for the others) the main priority is by the end of the year to have a robust, stable solution which is easy for the physicists to understand and use. If that requires substantial refactoring of things that already works function-wise I think that's what we should do because it will pay off long-term.

TeresiaOlsson:
> > . Do you still have an example of the solution you had before?
> > It was at the very beginning it was a @Validate decorator on the constructor.

I was mostly interested in the other parts. You said there were four duplications if we accept this PR:

- model
-  constructor signature
- object field for repr and edition
- attachment process

Without pydantic there would be no model but what was the solution you had previously for the "object field for repr and edition" and "attachment process"? If I have understood it correctly you avoid the duplication at the moment because you use the pydantic model for those things?

TeresiaOlsson:
> I would really prefer that you work on more important stuff rather than fighting with pydantic:
> 
>     * pyaml device abstraction
> 
>     * diagnostic devices such as emittance, beamloss, TbT, ...
> 
>     * disabled device
> 
>     * tuning/measurement tools (learn pySC and implement BBA, LOCO, ...)
> 
>     * Expert application for config (something similar to Jive)
> 
>     * Unit conversion
> 
>     * Documentation

I'm not fighting with pydantic. For me this PR is work for several things on your list:

- Making it easier to implement devices since pydantic is no longer needed for that. No ConfigModel is needed anymore but you can just write the device and add a schema later when/if you want validation.

- Possibility to generate a JSON schema which includes lab specific devices. The goal of that is to use either existing GUI applications or develop our own which uses the schema. That would allow to manage the configuration separately without loading pyAML.

- Make the documentation easier to understand since a user who wants to create a quadrupole can just look at the Quadrupole class and find all required attributes there instead of also having to look at the ConfigModel.

JeanLucPons:
My @validate decorator was a very first solution that has been discarded.
We made other choices.
The hardware layer refurbishment/simplification is in progress.
Please work on the list I mentioned.
Thanks for understanding.

TeresiaOlsson:
I'm currently working on the documentation and finishing this PR. As I understood the outcome of the maintainer meeting I'm supposed to rebase this PR after the merge for the catalog was done and then when all the tests are working again, I will flag the PR ready for review and it can be evaluated if it should be merged or not.

I'm still thinking about the duplication issue. What is in the branch at the moment is the current best solution I have found after testing many options but I haven't given up on it yet. Maybe there is a better solution.

TeresiaOlsson:
> I managed to get something working. I don't know if it produce a usable schema. But at least there are definitely something that we can do with arbitrary type.
> 
> ```python
> from dataclasses import dataclass
> from typing import Any
> from pydantic_core import core_schema
> from pydantic import BaseModel, GetCoreSchemaHandler, GetJsonSchemaHandler, ConfigDict
> from pydantic.json_schema import JsonSchemaValue
> from pydantic_core import core_schema as cs
> 
> class MyClass1ConfigModel(BaseModel):
>     param1:str
>     param2:int
> 
> class MyClass1:
> 
>     def __init__(self,cfg:MyClass1ConfigModel):
>         self._cfg = cfg
>     
>     @classmethod
>     def __get_pydantic_json_schema__(
>         cls, core_schema: cs.CoreSchema, handler: GetJsonSchemaHandler
>     ) -> JsonSchemaValue:
>         # Map the schema of the config object
>         sch = MyClass1ConfigModel.__get_pydantic_core_schema__(MyClass1ConfigModel,handler)
>         print(sch)
>         return sch
>         
> class MyClass2ConfigModel(BaseModel):
>  
>     model_config = ConfigDict(arbitrary_types_allowed=True, extra="forbid")
> 
>     param:str
>     extern:MyClass1
> 
> class MyClass2:
> 
>     def __init__(self,cfg:MyClass2ConfigModel):
>         self._cfg = cfg
> 
> 
> mycfg = MyClass1ConfigModel(param1="Param1Value",param2=10)
> mc1 = MyClass1(mycfg)
> mycfg2 = MyClass2ConfigModel(param="Param2Value",extern=mc1)
> mc2 = MyClass2(mycfg2)
> print(mc2)
> 
> print(MyClass2ConfigModel.model_json_schema())
> ```

I took a look at what you had done since I was planning to try the same myself but didn't have time to do it yet. I think this is the way to implement it if you want an arbitrary type to work the same as if it was a type already supported by pydantic:

```python
from dataclasses import dataclass
from typing import Any
from pydantic_core import core_schema
from pydantic import BaseModel, GetCoreSchemaHandler, GetJsonSchemaHandler, ConfigDict
from pydantic.json_schema import JsonSchemaValue
from pydantic_core import core_schema as cs

class MyClass1ConfigModel(BaseModel):
    param1:str
    param2:int

class MyClass1:

    def __init__(self,cfg:MyClass1ConfigModel):
        self._cfg = cfg

    @classmethod
    def __get_pydantic_core_schema__(
        cls, source_type: Any, handler: GetCoreSchemaHandler
    ) -> cs.CoreSchema:
        sch = handler.generate_schema(MyClass1ConfigModel)
        return sch    
    
    # @classmethod
    # def __get_pydantic_json_schema__(
    #     cls, core_schema: cs.CoreSchema, handler: GetJsonSchemaHandler
    # ) -> JsonSchemaValue:
    #     # Map the schema of the config object
    #     sch = MyClass1ConfigModel.__get_pydantic_core_schema__(MyClass1ConfigModel,handler)
    #     print(sch)
    #     return sch
        
class MyClass2ConfigModel(BaseModel):
 
    model_config = ConfigDict(arbitrary_types_allowed=True, extra="forbid")

    param:str
    extern:MyClass1

class MyClass2:

    def __init__(self,cfg:MyClass2ConfigModel):
        self._cfg = cfg


mycfg = MyClass1ConfigModel(param1="Param1Value",param2=10)
#mc1 = MyClass1(mycfg)
mycfg2 = MyClass2ConfigModel(param="Param2Value",extern=mycfg)
mc2 = MyClass2(mycfg2)
print(mc2)

print(MyClass2ConfigModel.model_json_schema())
```

As I understand it `__get_pydantic_core_schema__` covers both validation and the JSON schema whereas `__get_pydantic_json_schema__` only covers the JSON schema. It could be that if this is implemented it is possible to put `arbitrary_types_allowed=False` since pydantic doesn't consider it an arbitrary type anymore but I haven't tested that yet.

I will try it more next week.

TeresiaOlsson:
I noticed when I did it that the input to the second class became the ConfigModel object of the first class instead of the object of the first class or I got a validation error. Not sure if that's good or bad.

But it could potentially be one way to remove the ConfigModels completely and have the class be it's own source of truth without having it inherit from BaseModel. Might be possible to write a `__get_pydantic_core_schema__` which derives the schema from the constructor signature. And if you don't want that method to always be in the class by default to have it added using a decorator.

ChatGPT suggests the following steps to avoid having to implement the validation logic yourself:

1. Inspect the `__init__`.
2. Create a temporary BaseModel from the signature.
3. Use that model's validator.
4. Instantiate the class with the validated values.

Sounds like an interesting option to me worth to explore further.

TeresiaOlsson:
ChatGPT even claims that pydantic supports `handler.generate_schema(cls.__init__)` and can derive the schema from the constructor itself. To be tested...

JeanLucPons:
> ChatGPT even claims that pydantic supports `handler.generate_schema(cls.__init__)` and can derive the schema from the constructor itself. To be tested...

That would be a great improvement !

When I played a bit with your PR, i noticed some strange error coming from pydantic claiming that a "type" field was missing and in fact it was "class_path". It would be nice to avoid any alias that will be confusing with error message handling.

The Catalog and hardware layer is not yet done as it will have impact almost everywhere. So we do it step by step. 
There are still significant work so you have time to find a solution for:
- repr
- shallow copy during attachment process (see bellow)
- getter/setter of config elements. 
- Error handling

To avoid duplication of the signature during the attachment process, you can shallow copy most of objects expect for virtual magnets that need to share the MagnetModel instance (take care with that but i can help).
 
To answer to @gupichon , this "confusing" shallow copy of object on each holder was done to allow symmetry in each holder and create new refs on them while preserving deep copy of underlying object suck as model and calibration data (that can be heavy). It is possible to break this symmetry by using "control_modes" and then your object will be instantiated only once in each specified holder.

TeresiaOlsson:
I will make a separate PR next week to explore this option. So far it feels promising to me.

gubaidulinvadim:
> We have a working solution without duplication and that handle repr in a simple and efficient manner. The only missing added value is the json schema but I would do in a different manner because writing an expert tools for pyaml configuration based on json schema will be a nightmare.
> 
> There are several problems with the actual proposal:
> 
>     * 4 duplications of the parameters (Model, constructor, fields for repr, attachment). This is too much.
> 
>     * Strong dependance to pydantic with an overriding of **get_union_of_schema** and **model_schema** that can end in same problem than overriding **get_pydantic_core_schema**
> 
>     * Apparently no way to handle same schema with same signature and a different name that will be a problem with pyaml_cs_oa
> 
> 
> However lots of good other points.... I agree... Personally, I can already build a full configuration file (or dict) from our lattice files. At ESRF we even already have device name in our lattice file but this is not a standard.

@JeanLucPons this is a draft, so we have to give Teresia a chance to find solutions to the problems you flagged here. I agree with you that we should not run  in circles reimplementing things that already work. However, in the case of configuration we were explicitly given feedback that it needs improvement and some rework is possible (already lots of new features like catalogs, regular expressions in the config file, etc.). It is the only part of the project that was flagged as a bit too complex after the workshop. So, we have to suffer a bit. This PR has some potentially nice features and some annoying things too. However, when it's a draft it's too hard to review because nothing works and there are too many changes. I remind you that with this PR @TeresiaOlsson is somewhat learning how everything is structured right now, which can be somewhat challenging. But afterwards it should be easier to contribute to different parts of the project. 

@TeresiaOlsson In general I like the idea of Jean Luc that some big PR should start small by changing for example only a part of the config (single element). This does make the review a bit more straighforward. In the case of configuration it's very hard to do (and too late) but maybe for some other proposals it would make sense. 

@JeanLucPons @TeresiaOlsson @gupichon @gupichon-soleil Maybe I'll ask the steering committee to put a (reasonable) deadline for freezing configuration (no large reworks, only improvements). That way it will be clear for everyone that we have to converge on a working solution in a reasonable time. What do you think? I can also give you the deadline myself :) Let's say the week of July 13th we freeze. I would say that we should also ask the steering committee that the next community meeting (September?) should be an online-hackathon where people should try the code and give their feedback.

JeanLucPons:
I quickly tested the creation of a dynamic model with arbitrary type and it worked.
So it is possible to a have single source of truth which is the constructor signature.

```python
from pydantic import create_model, ConfigDict
import inspect
import json

class MyClass1:
    
    def __init__(self,param1: str,param2: int):
        ...

class MyClass2:

    def __init__(self,param1:str,param2:MyClass1):
        ...

def register_schema(cls):

    # Inspect constructor signature and create dyammic schemas
    # for both validation and json schema export
    sig = inspect.signature(cls.__init__)
    fields = {}
    sch_fields = {}
    for param_name, param in sig.parameters.items():
        # Jump self
        if param_name == 'self':
            continue
        # Default value,
        if param.default != inspect.Parameter.empty:
            default = param.default
        else:
            default = ... # No default value specified -> Required field
        # Type
        annotation = param.annotation if param.annotation != inspect.Parameter.empty else str        
        sch_fields[param_name] = (annotation, default)
        fields[param_name] = (annotation, default)
        if annotation is MyClass1:
            # Arbitrary type example
            # Replace by its ConfigModel (for json schema export)
            annotation = MyClass1._model
        sch_fields[param_name] = (annotation, default)

    # Create dynamic model for the class
    model_name = cls.__module__ + "." + cls.__name__
    cls._model = create_model(model_name, __config__=ConfigDict(arbitrary_types_allowed=True), **fields)
    cls._schemamodel = create_model(model_name+"Schema", __config__=ConfigDict(arbitrary_types_allowed=True), **sch_fields)

    # Override __init__ to handle validation
    original_init = cls.__init__

    @classmethod
    def custom_init(cls, *args, **kwargs):        
        # Validate using dynamic model
        sig = inspect.signature(original_init)
        cls._model.model_validate(kwargs)
        original_init(cls, *args, **kwargs)

    cls.__init__ = custom_init
    return cls


register_schema(MyClass1)
register_schema(MyClass2)

mc1 = MyClass1(param1="test1",param2=10)
mc2 = MyClass2(param1="test2",param2=mc1)

top_level_json = MyClass2._schemamodel.model_json_schema()
with open('sch.json', 'w') as fp:
    json.dump(top_level_json, fp)
```

TeresiaOlsson:
Personally I don't see how we can freeze for refactoring. Part of the culture of open source is that everyone is allowed to make PRs about everything at any time if they have a suggestion. It doesn't mean the PR has to be accepted and merged in. But if you want the vague "before the summer vacations" deadline the steering committee has given the maintainers for simplification of configuration to mean the week of the 13th July that sounds like a good idea to make the deadline concrete.

For the community meetings the steering committee currently isn't so involved in it. They only approve a date that I set and then I come up with the agenda myself. If you think it would be a good idea to make it into a hackathon we can do that. The only thing I have lined up for a future meeting is a presentation about osprey but that doesn't have to be on the September meeting.

I agree that making smaller changes to merge things in in steps (like you are doing for the catalog) would be preferable. I will try to do that. Only issue is that we need to rewrite the tests since many of them currently relies on loading a configuration file. I think we should try to make unit tests which rely on as few other parts of the code as possible so parts can be tested individually. And then we can have integration tests which tests things together. I don't see the need to test loading the configuration in many tests.

gubaidulinvadim:
> Personally I don't see how we can freeze for refactoring. Part of the culture of open source is that everyone is allowed to make PRs about everything at any time if they have a suggestion. It doesn't mean the PR has to be accepted and merged in. But if you want the vague "before the summer vacations" deadline the steering committee has given the maintainers for simplification of configuration to mean the week of the 13th July that sounds like a good idea to make the deadline concrete.
> 
> For the community meetings the steering committee currently isn't so involved in it. They only approve a date that I set and then I come up with the agenda myself. If you think it would be a good idea to make it into a hackathon we can do that. The only thing I have lined up for a future meeting is a presentation about osprey but that doesn't have to be on the September meeting.
> 
> I agree that making smaller changes to merge things in in steps (like you are doing for the catalog) would be preferable. I will try to do that. Only issue is that we need to rewrite the tests since many of them currently relies on loading a configuration file. I think we should try to make unit tests which rely on as few other parts of the code as possible so parts can be tested individually. And then we can have integration tests which tests things together. I don't see the need to test loading the configuration in many tests.

There are no issues with PRs and changes for me. It's only "freezing" to keep us on track so we can meet all the milestones. I think the PR is timed well because it is now that we need to improve the configuration.

JeanLucPons:
I updated my code on the above post. I did 2 schema to handle validation and json schema.
There are definitely plenty of other ways to handle this with one single source of truth and arbitrary type.
We should then fall back in a situation similar to my [first proposal](https://github.com/orgs/python-accelerator-middle-layer/discussions/2#discussioncomment-13563256).

TeresiaOlsson:
> I quickly tested the creation of a dynamic model with arbitrary type and it worked. So it is possible to a have single source of truth which is the constructor signature.
> 
> ```python
> from pydantic import create_model, ConfigDict
> import inspect
> import json
> 
> class MyClass1:
>     
>     def __init__(self,param1: str,param2: int):
>         ...
> 
> class MyClass2:
> 
>     def __init__(self,param1:str,param2:MyClass1):
>         ...
> 
> def register_schema(cls):
> 
>     # Inspect constructor signature and create dyammic schemas
>     # for both validation and json schema export
>     sig = inspect.signature(cls.__init__)
>     fields = {}
>     sch_fields = {}
>     for param_name, param in sig.parameters.items():
>         # Jump self
>         if param_name == 'self':
>             continue
>         # Default value,
>         if param.default != inspect.Parameter.empty:
>             default = param.default
>         else:
>             default = ... # No default value specified -> Required field
>         # Type
>         annotation = param.annotation if param.annotation != inspect.Parameter.empty else str        
>         sch_fields[param_name] = (annotation, default)
>         fields[param_name] = (annotation, default)
>         if annotation is MyClass1:
>             # Arbitrary type example
>             # Replace by its ConfigModel (for json schema export)
>             annotation = MyClass1._model
>         sch_fields[param_name] = (annotation, default)
> 
>     # Create dynamic model for the class
>     model_name = cls.__module__ + "." + cls.__name__
>     cls._model = create_model(model_name, __config__=ConfigDict(arbitrary_types_allowed=True), **fields)
>     cls._schemamodel = create_model(model_name+"Schema", __config__=ConfigDict(arbitrary_types_allowed=True), **sch_fields)
> 
>     # Override __init__ to handle validation
>     original_init = cls.__init__
> 
>     @classmethod
>     def custom_init(cls, *args, **kwargs):        
>         # Validate using dynamic model
>         sig = inspect.signature(original_init)
>         cls._model.model_validate(kwargs)
>         original_init(cls, *args, **kwargs)
> 
>     cls.__init__ = custom_init
>     return cls
> 
> 
> register_schema(MyClass1)
> register_schema(MyClass2)
> 
> mc1 = MyClass1(param1="test1",param2=10)
> mc2 = MyClass2(param1="test2",param2=mc1)
> 
> top_level_json = MyClass2._schemamodel.model_json_schema()
> with open('sch.json', 'w') as fp:
>     json.dump(top_level_json, fp)
> ```

So the approach is to not have the `ConfigModel` but instead generate it automatically from the signature of the domain class as part of registering it in the schema registry? That sounds like an interesting idea and would also be possible to combine with the option to write your own dedicated schema class in case you want to use some of the more advanced validation features of pydantic like `model_validator` and `field_validator`. But I don't fully understand why you want to overwrite the constructor? Is it just because you want validation to happen automatically when creating the object instead of making the validation a separate step that is optional to do before?

I like this approach better than what I tested in #273 because that suddenly mixed validation and object creation and allowed pydantic to resolve nested models by itself and start to act like a factory. This approach doesn't seem to do that since you need to first create an object of `MyClass1` and give that to `MyClass2`. So the object creation is still separated from the validation.

TeresiaOlsson:
I think the difference compared to adding `__get_pydantic_core_schema__` is that `mc2 = MyClass2(param1="test2",param2=mc1)` will only validate so `mc1` is of the correct type when running `cls._model.model_validate(kwargs)` and not do validation of the fields inside `mc1`. But that is not a major problem for me since the `SchemaValidator` will anyway do that type of nested validation. I don't think we need pydantic to do it for us.

gupichon:
I'm not sure that overriding the constructor is a good idea. Model validation should not be done inside the class constructor, as you may want to validate data before the actual instantiation. This validation could be shared between the configuration validation layer and the constructor (as it might be used directly). Maybe in a class method with a decorator?

TeresiaOlsson:
I agree. But I think overriding the constructor is not needed. We only need that part if we want the class to validate automatically when an object is created. If not, we can keep the validation in a separate step using the SchemaValidator. That only requires the schema to be validated against to be registered in the SchemaRegistry.

gupichon:
> That only requires the schema to be validated against to be registered in the SchemaRegistry.

Yes, that's why I thinking about two decorators: one for registration and another for validation. If the second one is absent, it means no validation, which is a good first-step approach.

JeanLucPons:
Personally I would prefer to have the validation enabled by default for both "on the fly" construction and when loading.
Then a way to disable it globally by using the registry or for one single class.

JeanLucPons:
For configuration param getter/setter I imagine something like below.
The param1 property can be automatically defined to avoid supplication but , if we do like that, the code completion is lost in VS code :(

```python
class ConfigReadWriteStringScalar:
    def __init__(self,parent: object,parameter:str):
        self.parent = parent
        self.parameter = parameter
    def set(self,value:str):
        setattr(self.parent,self.parameter,value)
    def get(self):
        return getattr(self.parent,self.parameter)


def assign_param(object,frame):
    ai:inspect.ArgInfo = inspect.getargvalues(frame)
    for argname,value in ai.locals.items():
        if argname != "self":
            setattr(object,"_pyaml_"+argname,value) # We can define a syntax usable for repr as well

class MyClass1:
    
    def __init__(self,param1: str,param2: int):
        # Assign parameter to local fields
        # To use this helper function, the constructor cannot be overloaded as above (lost of signature)
        assign_param(self,inspect.currentframe())

    # Follow the pyaml coding style and code completion is OK
    @property
    def param1(self) -> ConfigReadWriteStringScalar:
        return ConfigReadWriteStringScalar(self,"_pyaml_param1")
```

TeresiaOlsson:
> Personally I would prefer to have the validation enabled by default for both "on the fly" construction and when loading. Then a way to disable it globally by using the registry or for one single class.

What do you think about separating it into two different decorators like @gupichon suggested? One which registers into the schema registry and one which adds the validation to the constructor? That feels easier to understand to me than one which is doing both things.

TeresiaOlsson:
> For configuration param getter/setter I imagine something like below. The param1 property can be automatically defined to avoid supplication but , if we do like that, the code completion is lost in VS code :(
> 
> ```python
> class ConfigReadWriteStringScalar:
>     def __init__(self,parent: object,parameter:str):
>         self.parent = parent
>         self.parameter = parameter
>     def set(self,value:str):
>         setattr(self.parent,self.parameter,value)
>     def get(self):
>         return getattr(self.parent,self.parameter)
> 
> 
> def assign_param(object,frame):
>     ai:inspect.ArgInfo = inspect.getargvalues(frame)
>     for argname,value in ai.locals.items():
>         if argname != "self":
>             setattr(object,"_pyaml_"+argname,value) # We can define a syntax usable for repr as well
> 
> class MyClass1:
>     
>     def __init__(self,param1: str,param2: int):
>         # Assign parameter to local fields
>         # To use this helper function, the constructor cannot be overloaded as above (lost of signature)
>         assign_param(self,inspect.currentframe())
> 
>     # Follow the pyaml coding style and code completion is OK
>     @property
>     def param1(self) -> ConfigReadWriteStringScalar:
>         return ConfigReadWriteStringScalar(self,"_pyaml_param1")
> ```

I'm not sure I understand the purpose of this. What is `ConfigReadWriteStringScalar?` for?

JeanLucPons:
> What do you think about separating it into two different decorators like @gupichon suggested? One which registers into the schema registry and one which adds the validation to the constructor? That feels easier to understand to me than one which is doing both things.

I'm sure to see how you want to implement it but overriding the constructor creates the issue I mentioned above + the potential lost of completion in IPython. The validation can be handled in the `assign_param` helper function instead.

> I'm not sure I understand the purpose of this. What is `ConfigReadWriteStringScalar?` for?

It allows to write:
```python
mc1 = MyClass1(param1="test1",param2=10)
mc2 = MyClass2(param1="test2",param2=mc1)
print(mc1.param1.get())
mc1.param1.set("new value")
print(mc1.param1.get())
```

gupichon:
Hi @TeresiaOlsson and @JeanLucPons,
Here is an example that offers more flexibility with a single decorator (maybe easier than using two, by leveraging decorator parameters).
I have added some examples.

```python
from pydantic import create_model, ConfigDict
from functools import wraps
import inspect
import json
import yaml


def register_schema(
    cls=None,
    *,
    override_constructor=True,
    custom_validation_method=None,
):
    def decorate(target_cls):
        # Inspect constructor signature and create dynamic schemas
        # for both validation and json schema export
        sig = inspect.signature(target_cls.__init__)
        fields = {}
        sch_fields = {}
        for param_name, param in sig.parameters.items():
            # Jump self
            if param_name == "self":
                continue
            # Default value,
            if param.default != inspect.Parameter.empty:
                default = param.default
            else:
                default = ...  # No default value specified -> Required field
            # Type
            annotation = (
                param.annotation if param.annotation != inspect.Parameter.empty else str
            )
            fields[param_name] = (annotation, default)

            # If another class has already been registered, use its schema model
            # for config validation and JSON Schema export.
            schema_annotation = getattr(annotation, "_schemamodel", annotation)
            sch_fields[param_name] = (schema_annotation, default)

        # Create dynamic model for the class
        model_name = target_cls.__module__ + "." + target_cls.__name__
        target_cls._model = create_model(
            model_name, __config__=ConfigDict(arbitrary_types_allowed=True), **fields
        )
        target_cls._schemamodel = create_model(
            model_name + "Schema",
            __config__=ConfigDict(arbitrary_types_allowed=True),
            **sch_fields,
        )

        def run_custom_validation(validated_config):
            if custom_validation_method is None:
                return
            custom_validation = getattr(target_cls, custom_validation_method)
            custom_validation(validated_config)

        def _validate_constructor_config(class_, config):
            validated_config = class_._model.model_validate(config)
            run_custom_validation(validated_config)
            return validated_config

        def validate_config(class_, config, *, input_format=None):
            if input_format not in (None, "json", "yaml"):
                raise ValueError("input_format must be None, 'json', or 'yaml'")

            if isinstance(config, str):
                if input_format == "json":
                    validated_config = class_._schemamodel.model_validate_json(config)
                elif input_format == "yaml":
                    validated_config = class_._schemamodel.model_validate(
                        yaml.safe_load(config)
                    )
                else:
                    try:
                        parsed_config = json.loads(config)
                    except json.JSONDecodeError:
                        parsed_config = yaml.safe_load(config)
                    validated_config = class_._schemamodel.model_validate(parsed_config)
            else:
                validated_config = class_._schemamodel.model_validate(config)

            run_custom_validation(validated_config)
            return validated_config

        target_cls._validate_constructor_config = classmethod(_validate_constructor_config)
        target_cls.validate_config = classmethod(validate_config)

        if override_constructor:
            # Override __init__ to handle validation
            original_init = target_cls.__init__
            original_sig = inspect.signature(original_init)

            @wraps(original_init)
            def custom_init(self, *args, **kwargs):
                # Validate using dynamic model
                bound_args = original_sig.bind_partial(self, *args, **kwargs)
                validation_data = {
                    name: value
                    for name, value in bound_args.arguments.items()
                    if name != "self"
                }
                target_cls._validate_constructor_config(validation_data)
                original_init(self, *args, **kwargs)

            target_cls.__init__ = custom_init
        return target_cls

    if cls is None:
        return decorate
    return decorate(cls)

@register_schema
class MyClass1:
    def __init__(self, param1: str, param2: int): ...


@register_schema(override_constructor=False, custom_validation_method="custom_validate")
class MyClass2:
    def __init__(self, param1: str, param2: MyClass1): ...

    @classmethod
    def custom_validate(cls, config):
        if config.param1 == config.param2.param1:
            raise ValueError("MyClass2.param1 must differ from MyClass1.param1")


@register_schema(override_constructor=False)
class MyClass3:
    def __init__(self, param1: str, param2: MyClass2): ...


# register_schema(MyClass1)
# register_schema(MyClass2)

mc1 = MyClass1(param1="test1", param2=10)
mc2 = MyClass2(param1="test2", param2=mc1)
mc3 = MyClass3(param1="test3", param2=mc2)

top_level_json = MyClass2._schemamodel.model_json_schema()
with open("sch.json", "w") as fp:
    json.dump(top_level_json, fp)


my_class1_json = """
{
  "param1": "from-json",
  "param2": 20
}
"""

validated_mc1 = MyClass1.validate_config(my_class1_json, input_format="json")
mc1_from_json = MyClass1(**validated_mc1.model_dump())


my_class2_yaml = """
param1: from-yaml
param2:
  param1: nested-from-yaml
  param2: 30
"""

validated_mc2 = MyClass2.validate_config(yaml.safe_load(my_class2_yaml))
validated_mc2_data = validated_mc2.model_dump()
mc2_param2 = MyClass1(**validated_mc2_data["param2"])
mc2_from_yaml = MyClass2(
    param1=validated_mc2_data["param1"],
    param2=mc2_param2,
)


my_class3_yaml = """
param1: top-from-yaml
param2:
  param1: nested-myclass2-from-yaml
  param2:
    param1: nested-myclass1-from-yaml
    param2: 40
"""

validated_mc3 = MyClass3.validate_config(my_class3_yaml, input_format="yaml")
validated_mc3_data = validated_mc3.model_dump()
mc3_param2_param2 = MyClass1(**validated_mc3_data["param2"]["param2"])
mc3_param2 = MyClass2(
    param1=validated_mc3_data["param2"]["param1"],
    param2=mc3_param2_param2,
)
mc3_from_yaml = MyClass3(
    param1=validated_mc3_data["param1"],
    param2=mc3_param2,
)
```

gupichon:
I've done a few tests and it's a bit trickier with inheritance...

TeresiaOlsson:
> > What do you think about separating it into two different decorators like @gupichon suggested? One which registers into the schema registry and one which adds the validation to the constructor? That feels easier to understand to me than one which is doing both things.
> 
> I'm sure to see how you want to implement it but overriding the constructor creates the issue I mentioned above + the potential lost of completion in IPython. The validation can be handled in the `assign_param` helper function instead.
> 
> > I'm not sure I understand the purpose of this. What is `ConfigReadWriteStringScalar?` for?
> 
> It allows to write:
> 
> ```python
> mc1 = MyClass1(param1="test1",param2=10)
> mc2 = MyClass2(param1="test2",param2=mc1)
> print(mc1.param1.get())
> mc1.param1.set("new value")
> print(mc1.param1.get())
> ```

My personal opinion is that it will be too difficult for the users to understand. They will forget to use `assign_param` and it is confusing that you input one type and then the property for that attribute needs to return another type if it is just something simple as a string. We need to remember that many of them are going to be new to python and their default way of writing their own custom devices is going to be to follow the normal python conventions for how to assign attributes and how they are used to writing classes in pyEPICS/pyTANGO.

The only way I see it working is by doing something similar to how it is done in `HAPPI` where every attribute is an object of a custom class which takes care of the functionality. I think the users can understand if they need to write something like this:

```python
class MyClass1:
    
    def __init__(self,param1: str, param2: int):
        self._param1 = pyAMLAttribute(param1)
        self._param2 = pyAMLAttribute(param2)

    @property
    def param1(self) -> pyAMLAttribute:
        return self._param1
```
and then the `pyAMLAttribute` will take care of the functionality. That I have seen people understanding since `ophyd-async` does something similar with signals. I think they understand it because the class makes it clearer that it is something which exist to provide some functionality. And it is much easier to teach people that every attribute of a class need to be of a specific type than to teach them that they need to write things a specific way.

But that is a different way to write classes which would affect the whole hardware abstraction layer. I'm not sure it's worth it compared to just doing the validation in a separate step before you initialize an object.

JeanLucPons:
> and then the `pyAMLAttribute` will take care of the functionality...

What you propose is OK for me.

TeresiaOlsson:
> I've done a few tests and it's a bit trickier with inheritance...

I think when using `create_model` to dynamically create the schema from the class the information about the inheritance must be kept or the JSON schema will not be able to resolve abstract classes into it's subclasses. Without that the JSON schema becomes useless. My current thinking is try to use virtual subclasses for it but not sure if it will work out.

TeresiaOlsson:
> > and then the `pyAMLAttribute` will take care of the functionality...
> 
> What you propose is OK for me.

In that case I would still leave it out of this PR. It already includes too many things and probably needs to be separated into steps somehow to be able to review and merge. One way might be to first implement the parts needed for the schema registry with as few changes as possible to the existing ConfigModels and classes (maybe even do it module by module if possible like you do for the catalog) and then in a second step remove the ConfigModels + add dynamic generation of them. Then in the third step implement the things needed for having the automatic validation when creating objects. What do you think?

I think there might be a way to make this PR smaller by using the same approach as for the dynamic generation to turn the ConfigModels into schema classes. Then they can be registered and the SchemaRegistry, SchemaValidator and SchemaGenerator can be reviewed and merged in without having to modify a lot of classes in the same PR. I'm working on that part now.

JeanLucPons:
> I think when using `create_model` to dynamically create the schema from the class the information about the inheritance must be kept or the JSON schema will not be able to resolve abstract classes into it's subclasses. Without that the JSON schema becomes useless. My current thinking is try to use virtual subclasses for it but not sure if it will work out.

Yes. in my first example I create 2 models for that. The idea is to rebuild dynamically exactly what you did manually in your PR to export json format but it cannot handle validation. You can update your PR, get rid off configmodel and build it dynamically but it will be a bit difficult to handle correctly the following part will all registered classes (including compound type).

```python
        if annotation is MyClass1:
            # Arbitrary type example
            # Replace by its ConfigModel (for json schema export)
            annotation = MyClass1._schemamodel  # sorry for the mistake (due to testing)
```

JeanLucPons:
> In that case I would still leave it out of this PR.

I would do it at same time and cut the PR into categories rather than features as we do for hardware abstraction. Otherwise you will face repr issue (and likely others).
You are changing something fundamental that has a strong impact everywhere.

gupichon:
I have noticed a major issue here. If you want to inherit from an existing class without adding any attributes, but simply want to change its behavior, it won't work. You would have to change the signature, which is quite annoying for high-level PyAML objects.

JeanLucPons:
Yes I think @TeresiaOlsson already mentioned this issue during the last meeting.
This is why she removed all magnet types (correctors, quad,...) and make them Magnet.
It will be an issue with pyaml_cs_oa but we will try to solve this issue.

TeresiaOlsson:
I just removed them because a JSON schema validator can't handle two schemas where everything is the same. In the eyes of the validator it is the same schema. So I thought that were no reason to maintain several classes. And my goal was also eventually to manage to have oneOf instead of anyOf in the schema but so far that has turned out to be difficult.

gupichon:
There is one way to do this by storing the registered schemas, but it's ugly and quite complex...

Also, if the schema is the same (inheritance without new attributes), it would be possible to have an optionnal attribute like `__class__` added by the decorator to specify that you want to use a specific version of the implementation, while using standard PyAML classes by default.

```yaml
param1: top-MyClass3
param2:
  __class__: MyClass2Special
  param1: nested-special
  param2:
    param1: nested-myclass1
    param2: 45
```

Actually, it's a bit like the "type" field we are using today.

JeanLucPons:
> ... to manage to have oneOf instead of anyOf in the schema but so far that has turned out to be difficult...

Is oneOf a part of the json pydantic schema ?

JeanLucPons:
> Is oneOf a part of the json pydantic schema ?

it is, at least, a part of JSON schema standard:
https://json-schema.org/draft/2020-12/json-schema-validation#section-7.3.7

TeresiaOlsson:
> > Is oneOf a part of the json pydantic schema ?
> 
> it is, at least, a part of JSON schema standard: https://json-schema.org/draft/2020-12/json-schema-validation#section-7.3.7

I think pydantic can give it but only if you use the discriminated unions with a discriminator field but I haven't tried. Mostly pydantic gives anyOf. Since we anyway need a custom JSON schema generator to replace the abstract classes in the schema with their implementation I just added to it to replace anyOf to oneOf. That was easy. The problem occurred when I tried to use it in the Metaconfigurator because it crashed and became weird when using oneOf instead of anyOf for some of the subschemas. 

However, I'm still trying to debug it because I'm not sure that the problem is with the schema. It could equally well be a bug on their side with nested models. In the generator I also added a Literal with the string of each class which I think should make the schemas unique and fulfill oneOf. But that had the same bugs.

I wanted oneOf to work because it renders much nicer in the GUI. Instead of being allowed to click several options, the user is only allowed to click one. For example, it is confusing for the user why they can choose something to be both a Quadrupole and a Sextupole at the same time. That only makes sense to someone who has looked into the details of JSON schema validation and know the difference between anyOf and oneOf.

PR #265, opened 2026-05-22
Mirror package structure in tests.

TeresiaOlsson:
I suggest to mirror the package structure in tests to make it easier to understand which tests that are connected to which subpackage and more easily distinguish unit tests from integration tests.

I think many of the tests are at the moment written as integration tests and should maybe be refactored to units tests to allow to test subpackages and module individually and then have dedicated integration tests. With a structure which makes it clearer to understand how the tests relate to the structure of the package that might be easier to do.


gubaidulinvadim:
I personally agree that this is more straightforward.

gubaidulinvadim (approved):

PR #266, opened 2026-06-01
Refactor factory

TeresiaOlsson:
I have made a suggestion for how to refactor the factory to make it more modular and easier to understand the different steps for new developers. I also added more detailed docstrings everywhere.

**Changes:**

- `depth_first_build` has been renamed to `build`.
- `build_unbound` has been moved to a method `instantiate` on `UnboundElement`.
- Added an `ElementRegistry` and moved the singleton behavior there since that seemed to be the only reason why the factory had to be a singleton.
- Remove the `_strategies` attribute since it currently isn't used in line with the YAGNI principle. It can be added back when needed.

### DIFF
--- a/pyaml/accelerator.py
+++ b/pyaml/accelerator.py
@@ -174,7 +174,7 @@ class Accelerator(object):
             cannot be created. pydantic schema that support that an
             object is not created should handle None fields.
         """
-        dev = Factory.depth_first_build(config, ignore_external)
+        dev = Factory.build(config, ignore_external)
         if not isinstance(dev, Element):
             raise PyAMLConfigException(
                 "Invalid device type, Element or sub classes of Element expected " + f"but got {dev.__class__.__name__}"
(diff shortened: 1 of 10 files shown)


JeanLucPons:
Looks ok for me thanks for code cleaning !

JeanLucPons (approved):

JeanLucPons:
Hi @TeresiaOlsson 
I forget something in the review:

https://github.com/python-accelerator-middle-layer/pyaml/blob/e89f2e51f8995f361b91d610c6170f335feb82f6/pyaml/configuration/factory.py#L392

Here, you make a deep copy of the dict wich may causes performance issue with large file.
Previously the dict was mutated for performance reason.

An other thing is the Lock in the `ElementRegistry`.
Locking like that make no sense for the purpose.
If you want to load several files in paralell the only way is to remove the singleton behavior of this registry.

Thanks to patch it.

TeresiaOlsson:
Okay. I will take a look at it tomorrow and make the changes.

PR #267, opened 2026-06-03
New catalog handling

JeanLucPons:
This PR fix catalog config for BESSY2 orbit.
(Note thata catalog are available only for BPM at the moment)

### DIFF
--- a/examples/BESSY2_example/BESSY2Orbit.yaml
+++ b/examples/BESSY2_example/BESSY2Orbit.yaml
@@ -10,11 +10,9 @@ controls:
   - type: pyaml_cs_oa.controlsystem
     prefix: "pons:"
     name: live
-    catalog: bpm-catalog
-catalogs:
-  - type: pyaml_cs_oa.epics_catalog
-    name: bpm-catalog
-    prefix: ""
+    catalog:
+      type: pyaml_cs_oa.dynamic_catalog
+      backend: epics
 data_folder: /data/store
 arrays:
   - type: pyaml.arrays.magnet


examples/BESSY2_example/BESSY2Orbit.yaml
```diff
@@ -10,11 +10,9 @@ controls:
   - type: pyaml_cs_oa.controlsystem
     prefix: "pons:"
     name: live
-    catalog: bpm-catalog
-catalogs:
-  - type: pyaml_cs_oa.epics_catalog
-    name: bpm-catalog
-    prefix: ""
```
gubaidulinvadim, line 17: Is this syntax not correct / outdated @gupichon-soleil ? How should prefix be handled if there are different prefixes for live and for a virtual-accelerator?

gupichon: Yes, as I understand it, @JeanLucPons  removed the catalogs from the accelerator level to put them at the control system level. This prevents the ability to share the same catalog instance between control systems (for example, a live one and a twin). To do so, you would have to write it into a separate file that you can refer to in several CS, but this file would then be loaded as many times as it is referenced.

JeanLucPons: Yes i removed the indirection in pyaml (catalog: field) and make catalog 100% part of the backend.
This is mandatory in order to simplify the ControlSystem abstraction.
The control system abstraction will be reduced to 2 abstract class (ControlSystem and DeviceAccess) with simple methods.

With existing backend, to share a same catalog in several control system you can use yaml link instead as shonw below.

```yaml
controls:
    tango_host: ebs-simu-1:10000
    name: twin
    catalog: catalogs/ebs_orbit_bpm_catalogs.yaml
    tango_host: acs:10000
    name: live
    catalog: catalogs/ebs_orbit_bpm_catalogs.yaml
```

gupichon: Yes, but as I explained, it duplicates the loading process. Moreover, it forces users to split their data into several files instead of leaving them the choice to do as they wish.

JeanLucPons: The duplication in the loading process can be easily optimized in the file loader process and i think this should be done for all yaml links.
The final goal is to be able to provide a simple abstraction as shown in tool such as badger, pySC pyTAC, ...
You should be able to write a ControlSystem backend in a couple of day without previous knowledge.
For instance, it took me half a day to write a Tango backend for pyTAC.

gubaidulinvadim: Does it really force users to split the data into several files @gupichon-soleil? At least @JeanLucPons ' example just has the same filename for two backends. If there's still a disagreement, you could open a new PR to discuss it there. I'll approve this one to fix the example for now.

gubaidulinvadim (approved):

examples/BESSY2_example/BESSY2Orbit.yaml
TeresiaOlsson: What about having the catalogs as their own category in the yaml file? I always thought about it as a database for devices with it's own configuration and that the prefix/tango host would be added to all the PVs/attributes in the catalog as part of initializing the control system so that part belongs to the controlsystem config and not the catalog configuration.

Then the catalog field in the controls isn't a yaml link but the name of which catalog to use. Maybe it is even possible to make it optional to specify a name in case you only have one catalog defined and it is clear that the same one should be used for all controls?

TeresiaOlsson: I would suggest something like this:

```
controls:
    tango_host: ebs-simu-1:10000
    name: twin
    catalog: bpm-catalog
    tango_host: acs:10000
    name: live
    catalog: bpm-catalog
catalogs:
  - type: tango.pyaml.tango_catalog
    name: bpm-catalog
    other configuration fields the catalog might need...
```

gupichon: @TeresiaOlsson, that was the original proposal. I'll let you take a look at the comment made by @JeanLucPons to understand why it was rejected. In short: to delegate it to the backend through a simple call (get_device()).

TeresiaOlsson: From @JeanLucPons comment I don't understand why it was rejected. Why is the format in the configuration file connected to how the backend is implemented? For me that is two separate concerns and it is the responsibility of the factory to put all the pieces together in the correct way.

Personally I think the original proposal makes the configuration file much easier to understand.

JeanLucPons: @TeresiaOlsson 

In the original proposal, the backend has some catalog related methods that are not mandatory and complexify the backend interface.

```python
    def set_catalog(self, catalog: Catalog | None):
        self._catalog = catalog

    def get_catalog(self) -> Catalog | None:
        return self._catalog

    @abstractmethod
    def get_catalog_config(self) -> Catalog | str | None:
        """Return backend catalog configuration for this control system, if any."""
        pass
```

A simple interface for the control system is mandatory. It can be reduced to one single `get_device() -> DeviceAcess` method on the `ControlSystem` class and the `DeviceAccess` interface itself. Even this reduced interface is more complex than the pyTAC one that only have only one ControlSystem class with 2 or 3 methods to implement.

At the end a minimal working backend should be able to fit in a single and simple python file.

gupichon: Basically, catalogs has become a backend feature. Devices and tools should only declare a simple reference that the backend will be fully in charge of handling. The catalogs are no longer in PyAML.

Therefore, PyAML now requires the control system to provide the device implementation for a given key. The fact that the backend is using a catalog is completely transparent to PyAML.

gupichon: I will look into what Taurus does to build its binding with Tango and EPICS. @patrickmadela explained to @gubaidulinvadim and me this morning how it works with URIs. It could be a very good solution to replicate in PyAML.

TeresiaOlsson: Implementation-wise the change makes sense. I'm just trying to understand why the format in the configuration file has to change? Is the point that the catalog is no longer configured by pyAML? It is supposed to be configured separately on it's own and then just used inside pyAML? Like as if it was an external database?

What is supposed to be inside the `catalogs/ebs_orbit_bpm_catalogs.yaml` file?

JeanLucPons: > What is supposed to be inside the `catalogs/ebs_orbit_bpm_catalogs.yaml` file?
Exactly what you had before. You can find the file in the tests directory.

If you have some field outside the `control` field, this impact necessarily pyaml-core and user that want build their own backend will have to deal with this additional complexity.

gupichon: What I'm not entirely happy with regarding this solution is that it complicates the possibility of separating the choice of a catalog (static, dynamic, custom, ...) from the choice of the backend. In my opinion, it delegates perhaps too many things to the CS backend. Basically, if you want to implement a catalog dedicated to your specific needs, you have to do it for a specific backend or make it work for a set of backends.

gubaidulinvadim: @TeresiaOlsson Config-file-wise, it's completely OK. It's a very small change in the config, as can be seen in this PR. For me, it's the same config-file-change level as your Schema proposal. The config file just reflects that the catalogue belongs to the control system and changes its type.

gupichon: Yes, I agree with @gubaidulinvadim. To me, it's not what I would have chosen, but we can live with it.

JeanLucPons: The Teseria [example](https://github.com/python-accelerator-middle-layer/pyaml/issues/269#issuecomment-4621736396
) is a good example showing that it is an error to try to implement all use cases in 1 or 2 backends.
So we need to have a simple interface. The refurbishment of the CS class in not yet done.

TeresiaOlsson: > > What is supposed to be inside the `catalogs/ebs_orbit_bpm_catalogs.yaml` file?
> > Exactly what you had before. You can find the file in the tests directory.
> 
> If you have some field outside the `control` field, this impact necessarily pyaml-core and user that want build their own backend will have to deal with this additional complexity.

Sorry, I don't understand. Which example should I look at? I looked at the BESSY II example and there it looks like this:
```
controls:
  - type: pyaml_cs_oa.controlsystem
    prefix: "pons:"
    name: live
    catalog:
      type: pyaml_cs_oa.dynamic_catalog
      backend: epics
```
so the yaml file for the catalog should only contain this?
```
type: pyaml_cs_oa.dynamic_catalog
backend: epics
```
But then the catalog doesn't seem to be used for anything since I still have the device configuration inside the pyAML configuration as before?

I then looked at the SOLEIL example. There is looks like this:
```
controls:
  name: live
  tango_host: localhost:11000
  catalog: bpm-catalog
catalogs:
- catalogs.yaml
```
I can see that in `catalogs.yaml` I have the device configuration in a new way but I thought this was the syntax that was removed?

TeresiaOlsson: I also don't understand what the catalog has to do with `pyaml_cs_oa` since that is just the package for the ophyd-async bindings. I thought the catalog was just going to contain information about which PV to read/set, units etc. Why is that dependent on which bindings I use? For me that is just data. If I want to switch bindings I need to write an entirely new catalog?

JeanLucPons: In your example a `DynamicCatalog` is used:
It allow to construct the PV direclty wihtout creating an additional layer.
```yaml
- type: pyaml.bpm.bpm
  name: BPMZ5D1R
  model:
    type: pyaml.bpm.bpm_simple_model
    x_pos: "ORBITCC:rdPos@0"
    y_pos: "ORBITCC:rdPos@1"
```

You can switch to StaticCatalog if you prefer as bellow:

```yaml
controls:
    tango_host: ebs-simu-3:10000
    name: live
    catalog: catalogs/ebs_orbit_bpm_catalogs.yaml
```

```yaml
type: tango.pyaml.static_catalog
name: bpm-catalog
entries:
  - type: tango.pyaml.static_catalog_entry
    key: srdiag/bpm/c01-01/SA_HPosition  # This string is free (you can have what you want). It refers to a BPM x_pos field
    device:
      type: tango.pyaml.attribute_read_only
      attribute: srdiag/bpm/c01-01/SA_HPosition
      unit: mm
```

TeresiaOlsson: Aha, so there are two different types of catalogs with different usage? Maybe @gubaidulinvadim can put on the agenda for the next maintainer meeting to go through the catalog implementation and usage? I feel like a walk-through and demonstration of the usage would be useful to be up-to-date with the latest version.

I want to start putting together a draft of the user documentation for how to use the new features that has been added to the configuration layer but I can't explain to others what I haven't first understood myself :/

gupichon: Yes, static catalogs are basically a naive implementation, mainly for development purposes. They are used in the tests.

JeanLucPons: There are no limitation on the number of catalog implementations, we can just propose few ones that will cover a non exhaustive part of the needs. We will never address all uses cases such as Tango pipe, structured PV, new DevDict tango type, etc....

I would expect to speak of:
- CS refurbishment plan
- Abstraction of PyAML devices to allow Ophyd devices for every PyAML device (Done today only for the tune monitor).
- Virtual accelerator on github (really needed and urgent, even a simplified version)

gubaidulinvadim: @TeresiaOlsson @JeanLucPons @gupichon Can you discuss catalogues in a separate discussion/PR? This is a merged PR, it will be impossible to find this discussion later if necessary. 

I'll put the suggested items on the agenda. Realistically, we could not discuss all of it next Friday. I would put priority on finishing the configuration-related discussions, catalogues and the CS refurbishment plan. For the virtual accelerator, we will see if it's ready next week or by the end of June for simple CI/CD. But the progress is rather good so far.

PR #270, opened 2026-06-04
Remove unnecessary locks in ElementRegistry.

TeresiaOlsson:
I have removed the unnecessary locks in the ElementRegistry for the factory.

For the dicts I have some questions:

- Isn't `result = dict(data)`just a shallow copy which should be better performance wise since it just copies the top-level keys? And we don't have so many top-level keys?
- If I should remove it for performance reasons there shouldn't I also remove it in `_strip_build_metadata`? There I do `cleaned = dict(data)`.

### DIFF
--- a/pyaml/configuration/factory.py
+++ b/pyaml/configuration/factory.py
@@ -27,18 +27,16 @@ class ElementRegistry:
     """
 
     _instance = None
-    _lock = Lock()
 
     def __new__(cls):
         """
         Return the singleton registry instance.
         """
 
-        with cls._lock:
-            if cls._instance is None:
-                cls._instance = super().__new__(cls)
-                cls._instance._elements = {}
-            return cls._instance
+        if cls._instance is None:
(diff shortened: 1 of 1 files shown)


JeanLucPons:
As it is a recursive call the shallow  copy happens at each level.
However the overhead of this shallow copy seems not significant ~4% (Tested on EBSOrbit.yaml)
I didn't check for the list but it should be similar.

JeanLucPons (approved):

TeresiaOlsson:
> As it is a recursive call the shallow copy happens at each level. However the overhead of this shallow copy seems not significant ~4% (Tested on EBSOrbit.yaml) I didn't check for the list but it should be similar.

Good. Then I leave the shallow copy for now since it feels safer in case someone tries to reuse the input dictionary. I think I even do that in an example when comparing the input to what is generated after validation. We can always remove it later if it turns out to be significant.

PR #280, opened 2026-06-11
Updated BESSY2 example

JeanLucPons:
This PR updates BESSY2 example.
This PR requires [#21](https://github.com/python-accelerator-middle-layer/pyaml-cs-oa/pull/21).


gupichon (approved):
I would like @TeresiaOlsson to have a look at it before merging. Just to be sure the grammar is clear.

PR #282, opened 2026-06-11
Fix device access list

JeanLucPons:
This PR fix wrong implementation of `DeviceAccessList` which was not performing well abstracted method checking.
This PR requires pyaml-cs-oa [fix](https://github.com/python-accelerator-middle-layer/pyaml-cs-oa/pull/22) that triggered the issue.
@gupichon I would like that you check carrefully the check the range checking which was not in unit test.
I cannot check myself while tango backend was not update.
Thanks

### DIFF
--- a/pyaml/control/abstract_impl.py
+++ b/pyaml/control/abstract_impl.py
@@ -95,21 +95,21 @@ def _iter_devices_and_ranges(devs: DeviceAccess | DeviceAccessList):
       - DeviceAccessList: yields N items based on get_devices() and get_range() flattening
     """
     # Single device
-    if hasattr(devs, "get") and hasattr(devs, "get_range") and not hasattr(devs, "get_devices"):
+    if isinstance(devs, DeviceAccess):
         r = devs.get_range()
         if r is None:
             r = [None, None]
         return [(devs, [r[0], r[1]])]
 
-    # Device list (expects get_devices() + get_range() flat list)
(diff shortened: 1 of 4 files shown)


gupichon:
@JeanLucPons, I will look into it. A unit test exists for ranges but it's a bit basic. (test_ranges_array.py)

JeanLucPons:
OK thanks.
In fact it is more difficult than it looks to implements correctly abstract list.

gupichon (dismissed):

JeanLucPons:
In fact I will have to remove the fact that `DeviceAccessList` is a list.
Otherwise the aggregator will have to be fully reconstructed each time a device is added.

gupichon:
Yes, but implementing the sequence the way you did should correct this, thanks to the delegation to the inner list, right?

JeanLucPons:
In fact it does not work because the Aggregator has to update its internal structure when an item is added or removed.
Or to reconstruct its internal structure at each list update.
Otherwise classic list behavior cannot be handled.
Today it is working because of the `add_devices` method but if you use the list normally, it fails.

JeanLucPons:
This time it should be OK.
The list behavior of `DeviceAccessList` has been removed due to potential complex internal struture of aggregator.
Only relevant method are present and the internal structure is not exposed. No longer risk to use the list if a bad way.

gupichon (approved):

PR #287, opened 2026-06-12
Updated CS example

JeanLucPons:
This PR update example/ControlSystem/tango.

It shows how a `BPM` or `DeviceAccess` can be accessed.
It is similar here to a dynamic catalog in pyaml-cs-oa or tango-pyaml.

```python

sr = Accelerator.load("config.yaml")

# print the BPM position
bpm = sr.live.get_bpm("BPM_C01-01") # bpm is a BPM
print(bpm.positions.get())

# Direct access to control system
da = sr.live.get_device("srdiag/bpm/c01-01/SA_HPosition") # da is a DeviceAccess
print(f"h={da.get()} {da.unit()}")
```

```
[ubuntu20acu.pons] > python main.py 
[-1.05921952e-05  3.91472705e-06]
h=-1.0592195175840235e-05 m
```

### DIFF
--- a/examples/ControlSystem/tango/main.py
+++ b/examples/ControlSystem/tango/main.py
@@ -1,4 +1,11 @@
 from pyaml.accelerator import Accelerator
 
 sr = Accelerator.load("config.yaml")
-print(sr.live.get_bpm("BPM_C01-01").positions.get())
+
+# print the BPM position
+bpm = sr.live.get_bpm("BPM_C01-01")  # bpm is a BPM
+print(bpm.positions.get())
+
+# Direct access to control system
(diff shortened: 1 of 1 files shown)


TeresiaOlsson:
I find that maybe the name `get_device` is a bit confusing because many users think device == element. Maybe it can be called `get_deviceaccess` to make it clearer what it is you will get or you think that name is too long?

JeanLucPons:
ok we can rename to get_device_access().
@gupichon-soleil what do you think ?
It will impact backends.

gupichon:
Yes, get_device_access() is ok. The change is minor.

PR #291, opened 2026-06-18
Move magnet model device to name (to be solved by catalog)

### LINKED ISSUE #290: Feature: Delegate the deviceaccess creation to the backends for all PyAML elements
Device access descriptions should be removed from the configuration files. Their creation should be delegated to the backends.

**Proposed solution**
Replace all device access elements with simple strings, and use the backend `get_device` method instead of the current PyAML resolution and attachment mechanism.

Alternatives have already been discussed.

**Example**
In the configuration files, this kind of element should be replaced by a simple string:

```yaml
    - type: tango.pyaml.attribute
      attribute: AN01-AR/EM-SX/SHF.01/strength
      unit: 1/m**2
````

For example, it could become:

```yaml
    - AN01-AR/EM-SX/SHF.01/strength
```

**Additional context**
The implementation is already done for several elements, such as BPMs and the betatron tune monitor.
Magnets are the main remaining work.

**Checklist**
* [x] I've assigned this issue to a project
* [x] I've @-mentioned relevant people

JeanLucPons:
This PR moves Magnet device access to new Catalog architecture in the backends.
Closes #290


JeanLucPons:
@ Guillaume
Fell free to update example.
I will take in charge unit test.

gupichon:
@JeanLucPons, I had to change a few lines of code to make the examples loadable, so I could make sure they are correct.

gupichon:
Converting this PR to draft since it's still a work in progress and tests aren't ready yet.
@JeanLucPons, Shall I let you finish it up? Feel free to mark it as ready for review when it's done.

JeanLucPons:
I would prefer to avoid any merge for main otherwise after a while git is not able to present correctly merge conflicts. Some old core can reappear.
I had lots of trouble in merging your catalog into main.
Please rebase instead and force push so that there are only one divergence from main.

JeanLucPons:
Done. Tested on both BESSY2 VA and EBS VA (orbit).

pyaml/magnet/serialized_magnet.py
```diff
@@ -171,4 +171,4 @@ def __repr__(self):
         return __pyaml_repr__(self)
 
     def get_devices(self) -> list[DeviceAccess]:
```
gupichon: This method returns a list of string now. It should be renamed get_device_names too.

gupichon (changes requested):

gupichon (dismissed):
It will be ok once the tests are corrected.

gupichon (approved):

PR #295, opened 2026-06-19
First integration test.

### LINKED ISSUE #294: Feature: Integration tests pipelines with beam simulator
## Description, motivation and use case

pyAML currently has unit tests for its core configuration and backend
abstractions, but it does not have an integration test proving that a real
Tango-backed digital twin can be started and consumed through the pyAML API.

The use case is to validate the integration between:

- pyAML `Accelerator` configuration loading;
- `tango-pyaml` as the Tango backend;
- the `dt4acc` SOLEIL twin running in Apptainer with its embedded Tango
  database.

A minimal first test should ensure that the dt4acc twin starts successfully,
that Tango devices are reachable, and that pyAML can instantiate an
`Accelerator` and read at least one live value from the twin.

## Proposed solution

Add an opt-in GitHub Actions integration workflow that:

1. Installs Apptainer.
2. Pulls the `dt4acc-soleil-twin` Apptainer image.
3. Starts the twin with the embedded Tango database.
4. Uses test data stored under `tests/integration/data`.
5. Waits until the twin reports that all Tango processes are ready.
6. Runs a pytest integration smoke test.
7. Uploads the dt4acc twin log as an artifact.

Add a minimal pytest test under `tests/integration` that:

- loads a small pyAML accelerator configuration;
- instantiates `pyaml.Accelerator`;
- connects to the live Tango backend through `tango-pyaml`;
- reads the RF reference frequency from the dt4acc `RingSimulatorDevice`;
- asserts that the value is finite and positive.

The test should be opt-in, for example via `PYAML_DT4ACC_INTEGRATION=1`, so
the normal unit-test workflow remains fast and independent from Apptainer/Tango.

## Describe alternatives you've considered

One alternative is to mock Tango or use the existing dummy Tango backend. This
is useful for unit tests, but it does not validate the real deployment path
involving Apptainer, PyTango, Tango DB startup, dt4acc device registration, and
live device reads.

Another alternative is to run only the dt4acc integration tests in the dt4acc
repository. That validates the twin itself, but not the pyAML side of the
contract: loading an `Accelerator`, resolving Tango catalog references, and
reading values through the pyAML abstraction.

A larger end-to-end test could also write magnet or RF setpoints. That should
be added later, but the first step should be a small smoke test to stabilize
the CI infrastructure.

## Example

Example test behavior:

```python
accelerator = Accelerator.load("tests/integration/pyaml_dt4acc_twin.yaml")

rf = accelerator.live.get_rf_plant("RF")
reference_frequency = rf.frequency.get()

assert reference_frequency > 0.0
```

Example workflow behavior:

```bash
apptainer run \
  --bind tests/integration/data:/data:ro \
  dt4acc-soleil-twin.sif \
  --force-kill-db \
  --tango-host 127.0.0.1:10000 \
  --accelerator-setup-file /data/accelerator_setup.json \
  --lattice-file /data/SOLEIL_II_V3631_sym1_V001_database.m

PYAML_DT4ACC_INTEGRATION=1 python -m pytest -v tests/integration
```

## Additional context

The dt4acc twin already provides an Apptainer image with embedded Tango
database support. The required input files should be stored in
`tests/integration/data` so the workflow is self-contained and reproducible.

The initial test should target a stable read-only value, such as
`simulator/ringsimulator/ringsimulator/reference_frequency`, to avoid changing
the twin state during CI.

## Checklist

gupichon:
## Description

This PR adds an opt-in dt4acc/Tango integration smoke test for pyAML.

The goal is to validate that a real Tango-backed dt4acc SOLEIL twin can be started from CI, consumed through the pyAML API, and used to read live values from a small FODO accelerator configuration.

The integration setup is based on:

- `pyaml` branch: `294-feature-integration-tests-pipelines-with-beam-simulator`
- `tango-pyaml` backend branch: `main` (`13b5673`, includes `get_device_access()` and backend catalog support)
- `pyaml-cs-oa` backend branch: `main` (`53edab6`, current base for upcoming OA backend integration data)

## Related Issue

- Fixes #294

- [x] new feature: added an opt-in GitHub Actions workflow that starts the dt4acc SOLEIL twin with Apptainer and runs a pyAML integration smoke test.
- [x] new feature: added a minimal FODO pyAML configuration and matching dt4acc input data under `tests/integration/data`.
- [x] bugfix: disabled the dummy `tango-pyaml` test backend when `PYAML_DT4ACC_INTEGRATION=1`, because the integration test must use the real Tango backend.
- [x] bugfix: the integration smoke test reads Tango readback values instead of setpoint values, because `tango-pyaml` `.get()` returns the last written setpoint.

- [x] Pytest configuration: when `PYAML_DT4ACC_INTEGRATION=1`, the default dummy control-system test packages are not automatically exposed, so the real installed backend packages are used.
- [x] CI workflow: added `.github/workflows/dt4acc-integration.yml`, which pulls/runs the dt4acc Apptainer image, waits for the RingSimulator heartbeat, runs the smoke test, and uploads the twin log artifact.
- [x] Pytest markers: added the `integration` marker for tests requiring an external runtime.
- [x] Test data: added a small FODO lattice, pyAML configuration, catalog file, and dt4acc accelerator setup data for the smoke test.

## Testing

The following tests compatible with pytest were added:
- [x] `tests/integration/test_dt4acc_twin_smoke.py`

The test:
- loads `tests/integration/data/fodo_1gev_6d_pyaml.yaml`;
- instantiates the pyAML accelerator;
- verifies the live Tango control system is available;
- reads the RF reference frequency through the configured catalog;
- reads a quadrupole magnetic-strength readback through the real Tango backend;
- asserts the read values are finite and positive.

Validation:
- [x] The dt4acc integration workflow passes on GitHub Actions.
- [x] The test remains opt-in locally unless `PYAML_DT4ACC_INTEGRATION=1` is set.


gupichon:
The test is running but it fails only because of the pyaml config file which is not up to date with the current pyaml version.

gupichon:
Hello @JeanLucPons, @TeresiaOlsson, @gubaidulinvadim, and @patrickmadela. The first version of the integration tests is ready with a simple lattice (thanks @GamelinAl).
The tests are very basic for now, but they will allow us to move forward on this basis.

JeanLucPons:
Hi,
Thanks for this simple FODO with sextu.
It would be nice to configure a tune and chroma tuning tool.

gupichon:
> Hi, Thanks for this simple FODO with sextu. It would be nice to configure a tune and chroma tuning tool.

Yes, but for now, I just want to validate this integration test implementation. Additional tests will be added in a future PR, and we can discuss the necessary tests then.

JeanLucPons:
Ok
Are the UUID strictly necessary ?
The writing will not be very convenient if i want to get a handle to a magnet.

gupichon:
> Ok Are the UUID strictly necessary ? The writing will not be very convenient if i want to get a handle to a magnet.

No, I would like to update the lattice to use better names. Once again, the purpose here is to set up the integration test mechanism. The data and tests can be updated later.

JeanLucPons:
IMHO, It is preferable to start with good dataset if possible.
I let you do as you want.

TeresiaOlsson:
Is the idea of the integration test that every time someone pushes to the repository it will run the test once for the TANGO version and once for the EPICS version? And then later also maybe for the DOOCS version?

I'm wondering if there is some way we can come up with a mock-up naming convention for the control system names that allows us to generate the yaml files for each control system as part of the test? Then we just need to maintain the lattice and that naming convention and not yaml files for each specific control system.

For EPICS the naming convention is very forgiving so maybe we can just generate EPICS PV names based on the TANGO attribute names?

gupichon:
> Is the idea of the integration test that every time someone pushes to the repository it will run the test once for the TANGO version and once for the EPICS version? And then later also maybe for the DOOCS version?

Yes, for now it's just TANGO, I need to work on the simulator to use EPICS properly.

> I'm wondering if there is some way we can come up with a mock-up naming convention for the control system names that allows us to generate the yaml files for each control system as part of the test? Then we just need to maintain the lattice and that naming convention and not yaml files for each specific control system.

It need to be evaluated. Actually, the files for integration teses are generated from the lattice.

> For EPICS the naming convention is very forgiving so maybe we can just generate EPICS PV names based on the TANGO attribute names?

We should look at it, yes.

TeresiaOlsson:
Sounds like a good idea then that we try it first for TANGO and when that is working we try to generate the EPICS version based on the same configuration.

At least for me the EPICS version is not urgent.

gupichon:
@JeanLucPons, @GamelinAl and I are looking for a more complete lattice. For example, the way to use FamName and UUID is not trivial. I'll try to come up with a better proposal tomorrow. In the meantime, I would like you to refocus the review on the test architecture.

gupichon:
> Sounds like a good idea then that we try it first for TANGO and when that is working we try to generate the EPICS version based on the same configuration.
> 
> At least for me the EPICS version is not urgent.

Yes, right now I just want a functional pipeline.

tests/integration/data/fodo_1gev_6d.m
```diff
@@ -0,0 +1,172 @@
+function ring = fodo_1gev_6d()
```
TeresiaOlsson, line 1: I'm wondering if we should start to move away from storing the lattice in a matlab format? It feels somewhat strange to me to develop a new middle layer with the purpose of moving away from matlab and still keep the lattice as m or mat files. I think we should only keep that for compatibility reasons and have the default format in pyAML be json using the save_json and load_json functionality in pyAT. Maybe this lattice is a good start for that move?

tests/integration/data/fodo_1gev_6d.m
```diff
@@ -0,0 +1,172 @@
...
+ring = {...
+    atringparam('FODO_1GeV_6D', 1000000000.0, 1, 'Particle', struct('name', 'relativistic', 'rest_energy', 0.0, 'charge', -1.0), 'beam_current', 0.0, 'cell_harmnumber', 90.0);...
+     atrfcavity('RF_001', 0.0, 300000.0, 351319286.71874994, 90, 1000000000.0, 'UUID', 'RF_001_cc245d84ca2b4ba9b54051069a4cfb8f');...
+   atquadrupole('QF_001', 0.4, 1.2580503144087112, 'UUID', 'QF_001_f2013521b99e4a76bed1db7ec67464b0');...
```
TeresiaOlsson, line 5: What is the purpose of the UUID field since it's not a field that normally exists in a pyAT lattice?

tests/integration/data/fodo_1gev_6d.m
```diff
@@ -0,0 +1,172 @@
...
+        atsbend('B_032', 1.0, 0.19634954084936207, 0.0, 'UUID', 'B_032_2d3c2be8fe5f41abb120ab3438a4ef66');...
+        atdrift('DR_032', 0.5);...
+};
+    function v=False()
+        v=false;
+    end
+    function v=True()
+        v=true;
+    end
+end
```
TeresiaOlsson, line 172: I suggest that the lattice should have enable_6D. I think the simulator by default should use 6D tracking and applications that require 4D tracking needs to handle turning that off and on again. If we use enable_6d in this lattice we will catch those errors.

TeresiaOlsson (commented):
I think we should try to reduce the number of yaml files we need for tests because it feels to me like they have started to become difficult to maintain. But I think that can be left for another future PR.

tests/integration/data/fodo_1gev_6d.m
gubaidulinvadim: It's internal SOLEIL thing leftover from how we do it (it exists because of MML), it will be removed/replaced.

gupichon: To reference the elements otherwise than with the index. The FamName is not unique.

tests/integration/data/fodo_1gev_6d.m
gupichon: @GamelinAl, @gubaidulinvadim I'll let you answer to this one.

tests/integration/data/fodo_1gev_6d.m
gupichon: I honestly don't have an opinion on this.

tests/integration/data/fodo_1gev_6d.m
JeanLucPons: I would prefer to enable 6D for the VA (if not done) this will avoid future limitation.
Concerning UUID, the FamName seems unique is this lattice so why UUID ?
And yes it is important to limit as much as possible config file number for unit test.
this starts to be difficult to maintain.

TeresiaOlsson: I would also skip the UUID and just use unique names in this lattice to make it as simple as possible.

TeresiaOlsson:
> I think we should try to reduce the number of yaml files we need for tests because it feels to me like they have started to become difficult to maintain. But I think that can be left for another future PR.

Actually, maybe over time we can use the yaml file for this lattice also for the unit tests that need to load a config file? 

And we can also use it for a tutorial in the documentation for how to set up a configuration like @gubaidulinvadim suggested. Then we can have an example for same lattice, same hardware and what you need to do depending on which control system you choose.

JeanLucPons:
> > I think we should try to reduce the number of yaml files we need for tests because it feels to me like they have started to become difficult to maintain. But I think that can be left for another future PR.
> 
> Actually, maybe over time we can use the yaml file for this lattice also for the unit tests that need to load a config file?
> 
> And we can also use it for a tutorial in the documentation for how to set up a configuration like @gubaidulinvadim suggested. Then we can have an example for same lattice, same hardware and what you need to do depending on which control system you choose.

I agree, but some BPM, steerers and other stuff should be added to this lattice to be able to cover all tests.
The goal is rather to have a simple lattice runnable in a Tango or EPICS VA to check backend.

tests/integration/data/fodo_1gev_6d.m
gubaidulinvadim: @gupichon, @TeresiaOlsson  is right, by default we can have 6D. 

@JeanLucPons, it will be removed. It's a leftover from the SOLEIL-specific scripts because the SOLEIL lattice does not have unique element names (a decision I do not support).

gubaidulinvadim:
> > > I think we should try to reduce the number of yaml files we need for tests because it feels to me like they have started to become difficult to maintain. But I think that can be left for another future PR.
> > 
> > 
> > Actually, maybe over time we can use the yaml file for this lattice also for the unit tests that need to load a config file?
> > And we can also use it for a tutorial in the documentation for how to set up a configuration like @gubaidulinvadim suggested. Then we can have an example for same lattice, same hardware and what you need to do depending on which control system you choose.
> 
> I agree, but some BPM, steerers and other stuff should be added to this lattice to be able to cover all tests. The goal is rather to have a simple lattice runnable in a Tango or EPICS VA to check backend.

This (BPM, dipolar correctors, etc. ) is planned next. Mainly, it needs to be included in the lattice, VA can already handle BPMs (w/o offset as of now) and correctors.  We already discussed with @GamelinAl  and @gupichon that this will be the next step after we see that the initial tests are working well on GitHub.

tests/integration/data/fodo_1gev_6d.m
gubaidulinvadim: @TeresiaOlsson .m is OK because it is a text file similar to json. .mat is less ok but it's still an HDF5-type. But overall, yes it's better to have .json as the standard format in the examples.

tests/integration/data/fodo_1gev_6d.m
GamelinAl: @gubaidulinvadim it's not to be removed :) ... And it was not existing in MML lattice but introduced recently for pyAML/Digital Twin lattices. 
As @gupichon says it's a unique reference, it's the way we do it at SOLEIL but it's not something we want to impose to other facilities.

tests/integration/data/fodo_1gev_6d.m
GamelinAl: @TeresiaOlsson you are fully right, I'll start making the next one using .json

tests/integration/data/fodo_1gev_6d.m
GamelinAl: It's all right to have unique names for me for tests, but it should not be a restriction to be able to use pyAML. I think there are at least a few historic lattices (for example SOLEIL one) which do not have unique names.

TeresiaOlsson: > It's all right to have unique names for me for tests, but it should not be a restriction to be able to use pyAML. I think there are at least a few historic lattices (for example SOLEIL one) which do not have unique names.

I agree. I know many labs that don't have unique names and it will block them from moving from MML to pyAML. Mostly not for technical reasons but for cultural ones because they are so used to thinking about their magnets as families without unique names. The uniqueness for them only comes from the device list in the MML configuration, but sometimes they don't even use that (for example MAX IV). If there is only one power supply for a family, they have just one device in the device list and haven't separated the configuration into the individual magnets. Same for new machine where there isn't a control system yet but you want to run simulated commissioning (like BESSY III).

I was hoping we could fix this problem in the lattice loader. So if you load a lattice where the names are not unique, pyAML will generate unique names for you internally.

gupichon:
@JeanLucPons, @TeresiaOlsson, I know there are limitations with the current tests. I am currently working on an improved version, but I would like to validate the test architecture with you first. It was a big and complicated task: a simulator is launched using Apptainer on GitHub within a specific pipeline, all without disturbing the unit tests. It wasn't easy to set up, so I'd like to validate this step first.

I promise I'll open a discussion soon about integration tests (which ones to include and on which data). However, to achieve that goal, I also need to work on other subjects first. For instance, I need to address the KickAngle parameters on a given magnet (correctors). This isn't possible right now in PyAML, but I need it for the next FODO lattice. Additionally, I would like to discuss creating an abstraction layer between pyAt elements and PyAML. There is a lot to talk about, so we won't do it all here. I will open issues and/or discussions for those.

Also, a JSON version of the lattice will be used.

To sum up, I didn't focus on the content of the tests themselves, but rather on the infrastructure required to run them. If you have any comments on this part, I would be very happy to hear them.

TeresiaOlsson:
> @JeanLucPons, @TeresiaOlsson, I know there are limitations with the current tests. I am currently working on an improved version, but I would like to validate the test architecture with you first. It was a big and complicated task: a simulator is launched using Apptainer on GitHub within a specific pipeline, all without disturbing the unit tests. It wasn't easy to set up, so I'd like to validate this step first.
> 
> I promise I'll open a discussion soon about integration tests (which ones to include and on which data). However, to achieve that goal, I also need to work on other subjects first. For instance, I need to address the KickAngle parameters on a given magnet (correctors). This isn't possible right now in PyAML, but I need it for the next FODO lattice. Additionally, I would like to discuss creating an abstraction layer between pyAt elements and PyAML. There is a lot to talk about, so we won't do it all here. I will open issues and/or discussions for those.
> 
> Also, a JSON version of the lattice will be used.
> 
> To sum up, I didn't focus on the content of the tests themselves, but rather on the infrastructure required to run them. If you have any comments on this part, I would be very happy to hear them.

Sounds like a good approach to me. I looked at the workflow and it looked good to me but I wasn't sure how to test it without pushing something that triggers the test. Is there a way to do that? If not, I think we should merge it in so we can start experience how well it works and what is missing.

gupichon:
Yes, the tests run on each commit of a PR. You can also run the pipeline manually on GitHub.

To run them locally, you'll need to start the simulator first. I should probably create a launcher to make this easier. Under Ubuntu, you can use Apptainer. For Windows, WSL is required or you'll have to clone the repository, but it's doable.

You can take a look at the twin here:
https://gitlab.synchrotron-soleil.fr/software-control-system/digitaltwin/dt4acc/dt4acc-soleil-twin
The README will guide you.

Then you can launch the tests individually with the env variable `PYAML_DT4ACC_INTEGRATION=1`.

JeanLucPons:
Hello,
I got issues.
Unfortunately i cannot run the twin on the same host that my dev machine.
So the TANGO_HOST to localhost is not compatible with my setup.

So i edited both config file and changed prefix or tango_host:

pyaml-cs-oa failed due to missing [patch](https://github.com/bluesky/ophyd-async/issues/1228#issuecomment-4621616207) ?!
```
E       TypeError: //ubuntu24acu:36025/simulator/ringsimulator/ringsimulator/reference_frequency has type <class 'float'>, expected None
```

I tried to disable pyaml-cs-oa by doing:
```
    ids=["tango-pyaml"],
```
it fails.
```
 Failed: In test_dt4acc_twin_accelerator_instantiates_and_reads_live_values: 2 parameter sets specified, with different number of ids: 1
```

What are this ids ? could the fixture param be commented ? Thanks
It is not obvious to retrieve the DB port from the twin due to high speed logging.

image

JeanLucPons:
Would it be possible to construct the config using fragment ?
It would avoid duplication of the config file and allow to change dynamically the tango host using an env variable.

gupichon:
> What are this ids ? could the fixture param be commented ? Thanks

Actually, IDs are just labels for each test sample. You should also comment out the 'pyaml-oa' file line.
This should work:

```python
@pytest.mark.parametrize(
    "config_file",
    [
        Path(__file__).parent / "data" / "fodo_1gev_6d_pyaml.yaml",
        # Path(__file__).parent / "data" / "fodo_1gev_6d_pyaml-oa.yaml",
    ],
    ids=[
        "tango-pyaml",
        # "pyaml-cs-oa"
    ],
)
```

gupichon:
> pyaml-cs-oa failed due to missing [patch](https://github.com/bluesky/ophyd-async/issues/1228#issuecomment-4621616207) ?!

It's weird, I can't reproduce it either on GitHub or locally on my laptop.

gupichon:
> Would it be possible to construct the config using fragment ?
> It would avoid duplication of the config file and allow to change dynamically the tango host using an env variable.

I'll do it when I'm able to write proper tests in another PR.

JeanLucPons:
OK for ophyd asynch i had to uninstall re install.
It is difficult for me to run the twin.
Why not in this PR ?
Why not starting with good architecture ?

gupichon:
> Why not in this PR ?
> Why not starting with good architecture ?

This is because, for certain tests, I would like to implement some new features in PyAML. Furthermore, the tests to be conducted and the data to be used will be open for discussion. I am currently writing several proposals.

Additionally, there is a clear distinction between the test technical architecture and the tests themselves. I would like to have this ready before we start working on the tests. Once this is ready, we'll be able to work on integration tests across several PRs in parallel without major merge conflicts.

TeresiaOlsson:
My feeling is that we for both the TANGO and EPICS version of the twin could use a way to dynamically generate the host/prefix based on something user specific.

For the EPICS twin the prefix is already based on the username which so far has worked well for us. For TANGO maybe the port can be set based on the user ID + some number? That's how we generate unique ports for each user in our JupyterHub setup and that has also worked well so far.

And then we need to add the option in the configuration that you can specify an environment variable for the host/prefix so you don't have to manually change the yaml file because that people already complained about at the workshop.

JeanLucPons:
I just ask to be able to run the twin more easily.
This is the basis of this PR and not link with any advanced tests.
Just to read 1 or 2 attribute from the twin.

gupichon:
> And then we need to add the option in the configuration that you can specify an environment variable for the host/prefix so you don't have to manually change the yaml file because that people already complained about at the workshop.

Yes, I'll remove it from the file and set it at the pipeline level.

TeresiaOlsson:
Maybe there are two parts in this PR?:

1. The GitHub workflow for automatically running integration tests using virtual accelerator.

2. A new TANGO twin that also could be used independently of running any test. Also for developers and new users to run on their own to try some things out.

Sounds to me like 1 works well but 2 not because @JeanLucPons runs into the same issues we also had at the workshop where it needs to run in different environments.

But maybe the issues related to 2 is better to move to a separate PR? Then the integration test where it only needs to work for the GitHub workflow environment can be solved first.

gupichon:
> I just ask to be able to run the twin more easily. This is the basis of this PR and not link with any advanced tests. Just to read 1 or 2 attribute from the twin.

Alright, I'm writing a Linux launcher. It will be Linux-only, and I'm placing it right next to the test. It will be completely PyAML-independent and executable anywhere. It will only take the requieredconfiguration files as parameters.

JeanLucPons:
My main problem is that to run the integration test locally i have to change each time the 2 config files which is not convenient at all. This is why i propose a solution to avoid any prefix (tango host) definition in tests.
A second problem is that it is not trivial to extract the tango host from the twin due to logging.

gubaidulinvadim:
> OK for ophyd asynch i had to uninstall re install. It is difficult for me to run the twin. Why not in this PR? Why not start with good architecture ?

I think @gupichon  is trying to follow your principle (a good one) of doing small PRs :) The architecture of the test is to be improved (in this PR). It is true that running the twin locally is not very optimal now (to be fixed). 

Extensive logging will be fixed too. We already gave this feedback to Waheed (and there are fewer logs now than there were before, but still too many).

gupichon:
> My main problem is that to run the integration test locally i have to change each time the 2 config files which is not convenient at all.

It's corrected.

JeanLucPons:
> I think @gupichon is trying to follow your principle (a good one) of doing small PRs :) The architecture of the test is to be improved (in this PR). It is true that running the twin locally is not very optimal now (to be fixed).
> 

What I propose is to reduce the PR by avoiding config duplication and allow dynamic tango host change.

> Extensive logging will be fixed too. We already gave this feedback to Waheed (and there are fewer logs now than there were before, but still too many).

Good

gupichon:
> A second problem is that it is not trivial to extract the tango host from the twin due to logging.

As I said, I will write a great script that mimics what the pipeline does, featuring clean logging messages with the full TANGO host and a 'ready to use' notification.

JeanLucPons:
> It's corrected.

If i understand well your patch, it will then rely on TANGO_HOST env variable which is not good for us and that will fail with EPICS.

TeresiaOlsson:
> My main problem is that to run the integration test locally i have to change each time the 2 config files which is not convenient at all. This is why i propose a solution to avoid any prefix (tango host) definition in tests. A second problem is that it is not trivial to extract the tango host from the twin due to logging.

I agree. I have experienced the same problems with the EPICS twin. I will check with @Sulimankhail what he thinks about the option to dynamically generate the tango host similarly to how it's done for the EPICS twin. But if that feature is added it might affect what the host becomes when using it in a Github workflow. Maybe for the workflow we still want the option to have a fixed port?

JeanLucPons:
As it is a CI, the solution should work straightforward for all use case (github runner) and local test.
Otherwise, it will be a nightmare.

Sulimankhail:
> > OK for ophyd asynch i had to uninstall re install. It is difficult for me to run the twin. Why not in this PR? Why not start with good architecture ?
> 
> I think @gupichon is trying to follow your principle (a good one) of doing small PRs :) The architecture of the test is to be improved (in this PR). It is true that running the twin locally is not very optimal now (to be fixed).
> 
> Extensive logging will be fixed too. We already gave this feedback to Waheed (and there are fewer logs now than there were before, but still too many).

Noted

gupichon:
> > It's corrected.
> 
> If i understand well your patch, it will then rely on TANGO_HOST env variable which is not good for us and that will fail with EPICS.

How do you want to do it then?
For EPICS, I'm still working on the simulator side, but I lack time, to be honest.

JeanLucPons:
> How do you want to do it then? For EPICS, I'm still working on the simulator side, but I lack time, to be honest.

Using fragment (ConfigurationManager) with dynamic config for the CS.
It will also allow to avoid the duplication of `fodo_1gev_6d_pyaml-oa.yaml` and `fodo_1gev_6d_pyaml.yaml`.

TeresiaOlsson:
> > > It's corrected.
> > 
> > 
> > If i understand well your patch, it will then rely on TANGO_HOST env variable which is not good for us and that will fail with EPICS.
> 
> How do you want to do it then? For EPICS, I'm still working on the simulator side, but I lack time, to be honest.

I think the idea is that the twin will generate it's own port based on the environment it is started in. Like how for the EPICS twin it reads the username from the system and puts that as the prefix for all the PVs.

And then we need to add in the pyAML configuration loader that it understands how to resolve environment variables. So you for example for the TANGO_HOST field can write 127.0.0.1:$UID and it will resolve what the environment variable is.

gupichon:
> I think the idea is that the twin will generate it's own port based on the environment it is started in. Like how for the EPICS twin it reads the username from the system and puts that as the prefix for all the PVs.

That is actually how the twin works. I'm just forcing it for the integration test.

> And then we need to add in the pyAML configuration loader that it understands how to resolve environment variables. So you for example for the TANGO_HOST field can write 127.0.0.1:$UID and it will resolve what the environment variable is.

The TANGO host and EPICS prefix are handled by the backends, so I'm not quite sure I follow what you are proposing here.

gupichon:
> Using fragment (ConfigurationManager) with dynamic config for the CS.
> It will also allow to avoid the duplication of `fodo_1gev_6d_pyaml-oa.yaml` and `fodo_1gev_6d_pyaml.yaml`.

I'll do that in the next PR with the new tests. In the meantime, setting the environment variable in the pipeline is a good approach anyway for the scope of this PR.

TeresiaOlsson:
> > I think the idea is that the twin will generate it's own port based on the environment it is started in. Like how for the EPICS twin it reads the username from the system and puts that as the prefix for all the PVs.
> 
> That is actually how the twin works. I'm just forcing it for the integration test.
> 
> > And then we need to add in the pyAML configuration loader that it understands how to resolve environment variables. So you for example for the TANGO_HOST field can write 127.0.0.1:$UID and it will resolve what the environment variable is.
> 
> The TANGO host and EPICS prefix are handled by the backends, so I'm not quite sure I follow what you are proposing here.

So when you start the TANGO twin locally it will already have a user-specific port?

My idea was just that when you give the host and prefix in the yaml file part of the string can be an environment variable which will be resolved when loading in the configuration. For the BESSY II examples I for example always need to manually switch `pons` to my username on my computer before I can run the example. Everyone at the workshop had to do the same and they complained about it. Instead in the yaml file you can write $USER (or some other syntax to indicate that it's an environment variable) and it will be resolved during runtime.

gupichon:
> So when you start the TANGO twin locally it will already have a user-specific port?

You can set it as a parameter. If you don't, it will use the first available port on localhost.

> My idea was just that when you give the host and prefix in the yaml file part of the string can be an environment variable which will be resolved when loading in the configuration. For the BESSY II examples I for example always need to manually switch `pons` to my username on my computer before I can run the example. Everyone at the workshop had to do the same and they complained about it. Instead in the yaml file you can write $USER (or some other syntax to indicate that it's an environment variable) and it will be resolved during runtime.

I understand now, but this needs to be done in the pyaml-cs-oa project. Or maybe in pyaml, if we want to introduce a specific grammar detection like we do for wildcards or filenames. Either way, it needs to be addressed in a separate issue.

gupichon:
> My idea was just that when you give the host and prefix in the yaml file part of the string can be an environment variable which will be resolved when loading in the configuration. For the BESSY II examples I for example always need to manually switch `pons` to my username on my computer before I can run the example. Everyone at the workshop had to do the same and they complained about it. Instead in the yaml file you can write $USER (or some other syntax to indicate that it's an environment variable) and it will be resolved during runtime.

By the way, it's a very good idea.

TeresiaOlsson:
> > My idea was just that when you give the host and prefix in the yaml file part of the string can be an environment variable which will be resolved when loading in the configuration. For the BESSY II examples I for example always need to manually switch `pons` to my username on my computer before I can run the example. Everyone at the workshop had to do the same and they complained about it. Instead in the yaml file you can write $USER (or some other syntax to indicate that it's an environment variable) and it will be resolved during runtime.
> 
> By the way, it's a very good idea.

Actually, it's not really my idea... I think it was part of the feedback from the workshop. It needs to be added to pyaml but no one had time to do it yet. I can try to do it after I have finished the PR for the schema registry.

JeanLucPons:
> > Using fragment (ConfigurationManager) with dynamic config for the CS.
> > It will also allow to avoid the duplication of `fodo_1gev_6d_pyaml-oa.yaml` and `fodo_1gev_6d_pyaml.yaml`.
> 
> I'll do that in the next PR with the new tests. In the meantime, setting the environment variable in the pipeline is a good approach anyway for the scope of this PR.

Why you don't wan't to reduce this PR ?
And there's no need to modify either pyaml-cs-ca or tango-pyaml to have dynamic prefix or tango-host.

gupichon:
> Why you don't wan't to reduce this PR ?

I already answered this point twice, but you might have missed it.
You can find it here:
https://github.com/python-accelerator-middle-layer/pyaml/pull/295#issuecomment-4780910317
And here:
https://github.com/python-accelerator-middle-layer/pyaml/pull/295#issuecomment-4787447196
I don't want to spend any more time on this since it will be removed anyway.

> And there's no need to modify either pyaml-cs-ca or tango-pyaml to have dynamic prefix or tango-host.

How do you want to do it then? With the ConfigurationManager? I think what @TeresiaOlsson wants is to extract the value from the environment.

JeanLucPons:
>  I don't want to spend any more time on this since it will be removed anyway.

OK do as you want. May be you prefer to have duplicate code because it is easier.
I spent much time in code reviews and trying to factor config files and you come with new duplicates.
That's mean that for the next PR the review will be harder, removal of duplication plus new tests plus i don't know what.
I remove myself from the reviewers of this PR.

TeresiaOlsson (approved):
I think we merge it in so we can start using the workflow infrastructure. There are changes needed to be able to run the test locally but those are better to address in a separate PR.

gupichon:
@JeanLucPons, I made the script so the twin is easier to launch locally.

gupichon:
> I think we merge it in so we can start using the workflow infrastructure. There are changes needed to be able to run the test locally but those are better to address in a separate PR.

I created a script used both in the pipeline and to run the twin locally.

PR #303, opened 2026-06-26
Move functionality to handle root folder into one class.

TeresiaOlsson:
I have made some small changes to gather all the functionality to handle the root folder into one class.

The purpose is just to make it easier to for the users to understand which methods are available for this and distinguish them from other functionality in the fileloader module.

**Differences:**

- Instead of having to import `get_root_folder`, `set_root_folder` and `get_path` individually you now just can import `ROOT` and it will return a module level object for the current root folder.
- You then can do `ROOT.get()`, `ROOT.set()` and `ROOT.expand_path()` for the same functionality as before.

### DIFF
--- a/pyaml/configuration/__init__.py
+++ b/pyaml/configuration/__init__.py
@@ -3,13 +3,12 @@ PyAML configuration module
 """
 
 from .factory import Factory
-from .fileloader import get_root_folder, set_root_folder
+from .fileloader import ROOT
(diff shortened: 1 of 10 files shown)


TeresiaOlsson:
Oh no... Other changes were also added into this commit by vs code... I need to remove those changes.

TeresiaOlsson:
I have removed those changes. Only changes related to the root folder should be included now.

JeanLucPons (approved):
Ok for me

PR #304, opened 2026-06-26
Modified file loader for readability

### LINKED ISSUE #310: Feature: Use environment variables in configuration file
There are use cases for being able to include environment variables in the configuration file and have them resolved by the file loader.

The most important one at the moment is to be able to define the TANGO host or the EPICS prefix in examples and test without having to hard code a value which the user needs to manually change.

**Proposed solution**
I have looked at the [resolvers in OmegaConf](https://omegaconf.readthedocs.io/en/latest/custom_resolvers.html) and think there are useful parts there to take inspiration from which would allow us to add additional resolvers in the future.

We add module wide resolvers like this:

```
Resolver = Callable[[str, LoadContext | None], Any]
RESOLVERS: dict[str, Resolver] = {}
```

and then a small decorator to register a resolver:

```
def resolver(name: str):
    def decorate(func: Resolver) -> Resolver:
        RESOLVERS[name] = func
        return func
    return decorate
```

Resolvers can then be written like this:

```
@resolver("env")
def resolve_env(...)

@resolver("file")
def resolve_file(...):
```

Question is then how the syntax for the resolvers should look like in the configuration file? In OmegaConf they use the syntax `${env:USER}` to make it clear that it is something that should be resolved. I think this syntax works nicely because it makes it clear what the start and end is, allowing for example to write `localhost:${env:UID}` and use environment variables as part of a string.

In our case we can have:
-  `${env:USER}` 
-  `${file:sub_config.yaml}`
- future other options...

But we also currently have two other syntax:
- `file:sub_config.yaml`
- `sub_config.yaml`

I suggest to no longer support `file:sub_config.yaml` and replace it with `${file:sub_config.yaml}`, but keep supporting the option to just put the file without any prefix. It will then be resolved based on the file extension being in the list of supported format similar to how it's done at the moment.

I have considered to keep supporting the `file:sub_config.yaml` syntax, but I think that makes it confusing for the users because it makes it unclear why there is the `${file:sub_config.yaml}` option.

I have also considered to use different syntax for environment variables and files but I think long term that will cause difficulties and confusion if the number of resolvers grow. It would be better to have one generic syntax that works for all resolvers.

Still, I think we should keep supporting the syntax to just put the file and resolve based on extension since we currently have many configuration files which follows that approach.

**Checklist**

TeresiaOlsson:
I have modified the file loader to hopefully make it easier to understand since I forget between times when I looked at it. There is now a slightly different structure and more docstrings and comments to help remember how it works.

Biggest change is that I added a context manager to keep track of the stack of files. This I expect that normal users will likely not really understand (since I also forget and need to look up how context managers work every time) but I think it still helps to improve the readability. I also added some comments to help understand the steps it does.

I also changed the name `use_fast_loader` to `include_locations` to maybe make it easier to understand why there is a fast and slow loader and why you not always wants to go for the fast one. It still says in the docstring that including locations will make it slower.

I changed so the default is now to use the fast loader and the user needs to specify if they want to use the slow loader and include the location metadata.

### DIFF
--- a/examples/BESSY2_example/BESSY2Orbit.yaml
+++ b/examples/BESSY2_example/BESSY2Orbit.yaml
@@ -277,7 +277,7 @@ devices:
   vcorr_array_name: VCorr
   name: DEFAULT_ORBIT_CORRECTION
   singular_values: 16
-  response_matrix: file:orm.json
+  response_matrix: ${path:orm.json}
 - type: pyaml.rf.rf_plant
   name: RF
   masterclock: (MCLKHX251C:freq)[KHz]
--- a/examples/BESSY2_example/BESSY2Tune.yaml
+++ b/examples/BESSY2_example/BESSY2Tune.yaml
@@ -668,7 +668,7 @@ devices:
   name: DEFAULT_TUNE_CORRECTION
   quad_array_name: QForTune
   betatron_tune_name: BETATRON_TUNE
-  response_matrix: file:trm.json
+  response_matrix: ${path:trm.json}
 - type: pyaml.tuning_tools.tune_response_matrix
   name: DEFAULT_TUNE_RESPONSE_MATRIX
   quad_array_name: QForTune
--- a/examples/SOLEIL_examples/tuning_tools.yaml
+++ b/examples/SOLEIL_examples/tuning_tools.yaml
@@ -6,7 +6,7 @@
   name: DEFAULT_TUNE_CORRECTION
   quad_array_name: QCORR
   betatron_tune_name: BETATRON_TUNE
-  response_matrix: file:trm.json
+  response_matrix: ${path:trm.json}
 - type: pyaml.tuning_tools.tune_response_matrix
   name: DEFAULT_TUNE_RESPONSE_MATRIX
   quad_array_name: QCORR
@@ -40,4 +40,4 @@
   vcorr_array_name: VCORR
   name: DEFAULT_ORBIT_CORRECTION
   singular_values: 16
-  response_matrix: file:orm.json
+  response_matrix: ${path:file:orm.json}
--- a/pyaml/accelerator.py
+++ b/pyaml/accelerator.py
@@ -275,7 +275,7 @@ class Accelerator(object):
         return Factory.build(config_dict, ignore_external)
 
     @staticmethod
-    def load(filename: str, include_locations: bool = True, ignore_external=False) -> "Accelerator":
+    def load(filename: str, include_locations: bool = False, ignore_external=False) -> "Accelerator":
         """
         Load an accelerator from a config file.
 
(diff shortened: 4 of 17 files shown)


JeanLucPons:
As @gupichon is the author of this cycle detection I let him review.
Take care that there is no unit test for absolute path which was working.

TeresiaOlsson:
Good point. I should also look at the unit tests and maybe add some more to check that the behaviour has actually not changed.

JeanLucPons:
Thanks for clarifying this code, few comment lines can help a lot !

TeresiaOlsson:
I also discovered now that there already is a path to load without adding location metadata which is what I needed to finish #279 where the factory no longer need to handle validation errors. I just need to use the fast loader for the case when the user wants to skip validation :)

TeresiaOlsson:
I'm wondering if we should change the default behaviour and use the fast loader as default? Then the user normally gets the fast loader and only if they get an error and want more information to be able to debug they need to switch to the slow loader.

JeanLucPons:
Yes this is a common workaround. You load using fast loader and when it fails you reload using slow loader. This can be done in the Accelerator class.

TeresiaOlsson:
The PR to add the option to use environment variables solving issue #310 has now been merged into this PR.

gupichon:
Many thanks for these developments @TeresiaOlsson! From my point of view, it's a great improvement.
However, I'd like to have a small test for environment variables like this one:

``` python
@pytest.mark.parametrize(
    ("install_test_package", "config_file"),
    [
        (
            {"name": "tango-pyaml", "path": "tests/dummy_cs/tango-pyaml"},
            "tests/config/env_var_config.yaml"
        )],
    indirect=["install_test_package"],
)
def test_environment_variable(install_test_package, config_file:str):
    test_var = "Environment variable test"
    with patch.dict(os.environ, {"TEST_VAR": test_var}):
        acc = Accelerator.load(config_file)
    assert acc.get_description()==test_var
```

With this file:
```yaml
type: pyaml.accelerator
facility: ESRF
machine: sr
energy: 6e9
data_folder: /data/store
description: ${env:TEST_VAR}
devices: []
```

Regarding this specific tag (env), I'm wondering if it shouldn't be resolved by Pydantic somehow, so that it could be used either in a file or within a generically built dictionary. This way, the `from_dict` method would resolve it too. Maybe with a model validator like the ones used in PySC. However, this will be for later anyway.

gupichon:
My bad, I've found the tests with environment variables.

gupichon (approved):
Great job!
The only missing feature would be the ability to use environment variables without configuration files, but that's truly not mandatory.

TeresiaOlsson:
> Great job! The only missing feature would be the ability to use environment variables without configuration files, but that's truly not mandatory.

I also like that idea. It feels like it should be straight forward to implement that as part of the validation step after the PRs for the schema registry are done. I think it should be possible to add a custom validator in pydantic which can resolve an environment variable.

PR #306, opened 2026-06-26
Integration tests improvements

gupichon:
## Summary

This PR updates the DT4ACC integration setup to use the new public `dt4acc-soleil-twin` image and the updated embedded Tango launcher interface.

It also replaces the previous single-file FODO test configuration with a dedicated FODO 1 GeV 6D integration dataset, including the DT4ACC accelerator setup, lattice, pyAML device configuration, Tango catalog, and pyaml-cs-oa catalog.

## Changes

- Update the DT4ACC image URI to the new public registry:
  `oras://registry.synchrotron-soleil.fr/software_control_system_public/dt4acc-soleil-twin:latest`
- Launch the twin through the new `tango-db-embedded` command.
- Use the new FODO 1 GeV 6D DT4ACC setup and lattice files in CI.
- Install `ophyd-async[tango, ca, pva]` for integration coverage.
- Checkout the PR head repository/ref explicitly in the integration workflow.
- Add split pyAML configuration files for the FODO 1 GeV 6D integration setup.
- Exercise both `tango-pyaml` and `pyaml-cs-oa` backends against the same twin.
- Add integration coverage for:
  - accelerator instantiation and live readback
  - all declared magnet strength readbacks
  - example configuration loading
  - chromaticity measurement
- Mark simulator tune attributes as read-only where appropriate.
- Fix the `Accelerator.from_dict` docstring parameter type.

PR #311, opened 2026-07-01
Add resolvers in file loader

TeresiaOlsson:
This PR is for issue #310.

I have added the possibility to have different resolvers in the loader. Currently only two exists for `env` and `file`. The syntax for both of them is the same $`{resolver:value}`.

The old syntax `file:` has been removed and all test configuration files updated to instead use the new syntax.

It is still supported to put a file without any syntax and it will be resolved as long as it has the file extension is in the `ACCEPTED_SUFFIXES` list.

### DIFF
--- a/examples/BESSY2_example/BESSY2Orbit.yaml
+++ b/examples/BESSY2_example/BESSY2Orbit.yaml
@@ -277,7 +277,7 @@ devices:
   vcorr_array_name: VCorr
   name: DEFAULT_ORBIT_CORRECTION
   singular_values: 16
-  response_matrix: file:orm.json
+  response_matrix: ${path:orm.json}
 - type: pyaml.rf.rf_plant
   name: RF
   masterclock: (MCLKHX251C:freq)[KHz]
--- a/examples/BESSY2_example/BESSY2Tune.yaml
+++ b/examples/BESSY2_example/BESSY2Tune.yaml
@@ -668,7 +668,7 @@ devices:
   name: DEFAULT_TUNE_CORRECTION
   quad_array_name: QForTune
   betatron_tune_name: BETATRON_TUNE
-  response_matrix: file:trm.json
+  response_matrix: ${path:trm.json}
 - type: pyaml.tuning_tools.tune_response_matrix
   name: DEFAULT_TUNE_RESPONSE_MATRIX
   quad_array_name: QForTune
--- a/examples/SOLEIL_examples/tuning_tools.yaml
+++ b/examples/SOLEIL_examples/tuning_tools.yaml
@@ -6,7 +6,7 @@
   name: DEFAULT_TUNE_CORRECTION
   quad_array_name: QCORR
   betatron_tune_name: BETATRON_TUNE
-  response_matrix: file:trm.json
+  response_matrix: ${path:trm.json}
 - type: pyaml.tuning_tools.tune_response_matrix
   name: DEFAULT_TUNE_RESPONSE_MATRIX
   quad_array_name: QCORR
@@ -40,4 +40,4 @@
   vcorr_array_name: VCORR
   name: DEFAULT_ORBIT_CORRECTION
   singular_values: 16
-  response_matrix: file:orm.json
+  response_matrix: ${path:file:orm.json}
--- a/pyaml/configuration/fileloader.py
+++ b/pyaml/configuration/fileloader.py
@@ -1,10 +1,12 @@
 """PyAML configuration file loader."""
 
-import collections.abc
 import io
 import json
 import logging
+import os
+import re
 from abc import ABC, abstractmethod
+from collections.abc import Callable, Hashable
 from contextlib import contextmanager
 from dataclasses import dataclass, field
 from pathlib import Path
@@ -19,11 +21,13 @@ from .. import PyAMLException
 
 logger = logging.getLogger(__name__)
 
-ACCEPTED_SUFFIXES = (".yaml", ".yml", ".json")
-FILE_PREFIX = "file:"
 
 LOCATION_KEY = "__location__"
 FIELD_LOCATIONS_KEY = "__fieldlocations__"
+ACCEPTED_SUFFIXES = (".yaml", ".yml", ".json")
+RESOLVER_PATTERN = re.compile(r"\$\{([^{}]+)\}")
+
+FILE_PREFIX = "file:"  # Kept for compatibility reasons with other modules
 
 
 class RootFolder:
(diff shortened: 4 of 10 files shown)


TeresiaOlsson:
The tests are currently failing because many of the configuration files includes a reference to `tune_response.json` which doesn't exist in the repository at the moment. Should this file be added?

The reason why the tests didn't fail before is because the behaviour has changed slightly in this PR. Now for `${file:}` the file is loaded and the contents appended to the configuration. Previously it seems like `file:` only resolved the path but didn't load the file. Which behaviour is the preferred one?

TeresiaOlsson:
The `FILE_PREFIX ` is kept at the moment because it is used by the `restfetcher` and didn't entirely understand how that module works so I didn't dare to change it. It however looked to me like maybe the new resolvers could be used there too?

kparasch:
> The tests are currently failing because many of the configuration files includes a reference to `tune_response.json` which doesn't exist in the repository at the moment. Should this file be added?
> 
> The reason why the tests didn't fail before is because the behaviour has changed slightly in this PR. Now for `${file:}` the file is loaded and the contents appended to the configuration. Previously it seems like `file:` only resolved the path but didn't load the file. Which behaviour is the preferred one?

I think we should not need the file to exist or you end up with circular problems. 

To generate the tune response, you need a pyaml configuration. 
To load the pyaml configuration you need tune response.

And if you already have a tune_response saved, you shouldn't need to generate it each time.

TeresiaOlsson:
> > The tests are currently failing because many of the configuration files includes a reference to `tune_response.json` which doesn't exist in the repository at the moment. Should this file be added?
> > The reason why the tests didn't fail before is because the behaviour has changed slightly in this PR. Now for `${file:}` the file is loaded and the contents appended to the configuration. Previously it seems like `file:` only resolved the path but didn't load the file. Which behaviour is the preferred one?
> 
> I think we should not need the file to exist or you end up with circular problems.
> 
> To generate the tune response, you need a pyaml configuration. To load the pyaml configuration you need tune response.
> 
> And if you already have a tune_response saved, you shouldn't need to generate it each time.

Ah, I see what you mean. The configuration actually has two different type of files and they should be handled differently:

1. Files that contains some data but that data should not be loaded in as part of the configuration. The configuration should only contain the path to the file and loading the data will happen later during runtime.

2. Files that contains subsections of the configuration. These should be resolved during the loading.

My question is then why for the tuning tools the first category is using the `file:` syntax and not doing the same as for the excitation curves? For those there is an attribute `file` which contains the path. Why for the tuning tools isn't there just `response_matrix` which should point to a path? What is the purpose of including `file:` as part of the string?

kparasch:
> > > The tests are currently failing because many of the configuration files includes a reference to `tune_response.json` which doesn't exist in the repository at the moment. Should this file be added?
> > > The reason why the tests didn't fail before is because the behaviour has changed slightly in this PR. Now for `${file:}` the file is loaded and the contents appended to the configuration. Previously it seems like `file:` only resolved the path but didn't load the file. Which behaviour is the preferred one?
> > 
> > 
> > I think we should not need the file to exist or you end up with circular problems.
> > To generate the tune response, you need a pyaml configuration. To load the pyaml configuration you need tune response.
> > And if you already have a tune_response saved, you shouldn't need to generate it each time.
> 
> Ah, I see what you mean. The configuration actually has two different type of files and they should be handled differently:
> 
> 1. Files that contains some data but that data should not be loaded in as part of the configuration. The configuration should only contain the path to the file and loading the data will happen later during runtime.
> 2. Files that contains subsections of the configuration. These should be resolved during the loading.
> 
> My question is then why for the tuning tools the first category is using the `file:` syntax and not doing the same as for the excitation curves? For those there is an attribute `file` which contains the path. Why for the tuning tools isn't there just `response_matrix` which should point to a path? What is the purpose of including `file:` as part of the string?

Excitation curves could also be annotated with "file:", but there is was never a need to do it because the .csv extension cannot be mistaken for a configuration file.

I agree "file:" itself is a bad name. The idea was that "file:" represents the file-path itself and is not interpreted as a configuration file and the expansion is skipped. If we would put just the "response_matrix.json" path, then the file loader would interpret it as configuration, try to look for a "type: " entry and then fail to load because it doesn't have one. On this point I don't understand your implementation of the `resolve_file` function. The point would be to actually not load the file, while the function loads it.

Generally I like the concept of the resolvers. The present implementation also gives a very obvious way to configure pyaml using a database:
```
from pyaml.configuration.fileloader import resolver

@resolver('db')
def resolve_db(name):
    # fetch configuration from database entry
    return name

from pyaml.configuration.fileloader import RESOLVERS
print(RESOLVERS) # db is registered in the RESOLVERS
```

TeresiaOlsson:
Okay. Then I understand. I had totally misunderstood the purpose of `file:`. The purpose of it is just to add the root path to make an absolute path out of a relative one. And not to load anything.

Then I should have a separate resolver for resolving subconfigurations. For those the file should be loaded and the content appended.

At the moment I have mixed those two things into the same resolver.

TeresiaOlsson:
After some thinking, I have a new suggestion for the syntax of the resolvers:

- ${env:}: resolve an environment variable

- ${path:}: resolve a file path. This is the previous file:. It makes an absolute path of a relative path but doesn't load the file.

- ${file:}: resolve a file. This loads the file and appends the content. For the supported file extensions this resolver will also be used without having to explicitly write the prefix for convenience but I think implementing it as a resolver similar to the others is still a nice idea.

JeanLucPons:
> After some thinking, I have a new suggestion for the syntax of the resolvers:
> 
> * ${env:}: resolve an environment variable
> * ${path:}: resolve a file path. This is the previous file:. It makes an absolute path of a relative path but doesn't load the file.
> * ${file:}: resolve a file. This loads the file and appends the content. For the supported file extensions this resolver will also be used without having to explicitly write the prefix for convenience but I think implementing it as a resolver similar to the others is still a nice idea.

This is OK for me.

gubaidulinvadim:
@TeresiaOlsson I think the proposed change is good. Please also check the config files in the examples (BESSY, SOLEIL, ESRF, etc.). They might require changes and currently the tests do not check if the changes broke the examples.

gupichon:
> @TeresiaOlsson I think the proposed change is good. Please also check the config files in the examples (BESSY, SOLEIL, ESRF, etc.). They might require changes and currently the tests do not check if the changes broke the examples.

Yes, in the PR #306 I added a simple loading test for those configuration files, so the test will fail if they are not up to date.

TeresiaOlsson:
I have added a separate resolver for path so the tests are working now. I also updated the examples and removed the `FILE_PREFIX` since it's no longer used by the file loader. It is however used by the restfetcher and the solution I added there might be a bit temporary and could be done in a better way in the future.

gubaidulinvadim (approved):

PR #314, opened 2026-07-10
Implemented BBA

JeanLucPons:
This PR implements BBA.
There is no most effective steerer detection and quad location detection yet. For the time being this has to be configured by hand.

```python
    SR = sr.design
    bba = SR.get_bba("BBA-BPM_C04-03")

    # Add a misalignement
    SR.get_bpm("BPM_C04-03").offset.set([20e-6, -15e-6])
    # Measure offsets
    bba.measure()
    assert np.abs(bba.h_offset() - 20e-6) < 2e-6
    assert np.abs(bba.v_offset() + 15e-6) < 2e-6
```


kparasch (approved):
I think it's great!

The only thing I don't like but could maybe get addressed in a future PR is that with this way, there is one BBA tuning tool per BPM.

JeanLucPons:
Yes this is a simple implementation.
For more advanced BBA tool (including BPM and quad search) we can see later.

gubaidulinvadim:
From the example it seems that it only does a single quad-bpm pair. Is this correct?

JeanLucPons:
Yes no paralell BBA for the moment.

kparasch:
Yes. For reference, the pySC BBA implemented here is the classic "model-independent" or "bow-tie" BBA algorithm adapted from MML.

gubaidulinvadim:
Like @kparasch pointed out, maybe at the end we will want only the parallel version in the configuration. Otherwise, we'll get a very long BBA config with only the names different. 

@kparasch Maybe there will be some contribution from SOLEIL to pySC with either fast BBA or fast ORM. To be checked (Sami will leave SOLEIL; maybe his replacement will be less inclined to only test the code). Or I'll do it because for SOLEIL II we do not plan to rely on the classic algorithm (takes too long).

JeanLucPons:
Yes here you can see the whole BBA config for EBS (using only pySC)
https://github.com/kparasch/ebs2pySC/blob/main/bba_config_tango.json

To decrease this config size, a search of quad and steerer for a given bpm is needed.
This will need some more complex modif (typically to get s position).

PR #330, opened 2026-07-10
Changes for schema registry for BPM

### LINKED ISSUE #316: Remove ConfigModel from BPM
This issue is for the BPM subpackage. To do:

- Remove ConfigModel and move to constructor of BPM instead

TeresiaOlsson:
Changes required for the schema registry and removing the config module for BPM.

### DIFF
--- a/.github/workflows/ruff_formatting.yml
+++ b/.github/workflows/ruff_formatting.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches: [ main ]
   pull_request:
-    branches: [ main ]
 
 jobs:
   ruff-check:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,7 +7,6 @@ on:
   push:
     branches: [ "main" ]
   pull_request:
-    branches: [ "main" ]
 
 permissions:
   contents: read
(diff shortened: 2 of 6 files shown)


TeresiaOlsson:
I'm wondering about the new docstring for the BPM. They were based on the docstrings for the ConfigModel but they say that all attributes should be catalog keys. But what if you are not using a catalog?

JeanLucPons:
```python
SR = sr.design
print( SR.get_bpm("BPM_C04-04") )
```

before:
```
BPM(peer='Simulator:design', name='BPM_C04-04', description=None, lattice_names=None, x_pos='srdiag/bpm/c04-04/SA_HPosition', y_pos='srdiag/bpm/c04-04/SA_VPosition', x_offset=None, y_offset=None, tilt=None)
```
after:
```
BPM(name='BPM_C04-04')
```

gubaidulinvadim:
@JeanLucPons which configuration file have you used to check that? I'm having trouble with reviewing Teresia's merge request as none of the config files in examples (SOLEIL or BESSY) seems to load the accelerator. (including EBSorbit.yaml from tests/)

JeanLucPons:
I used the simple program example that I mentioned that loads "tests/config/EBSOrbit.yaml".
The problem is that some information are lost in  `__repr__`.
It is a choice from Teresia.
Typically the peer is interesting because you know to what your object is attached.
You can also filter by fields in ElementArray.
But now, there is only the name.

gubaidulinvadim (approved):
I've made an issue #345 for string representation with the new configuration and I will try to find a solution to get string representation close to what it was before without loss of information.

PR #331, opened 2026-07-10
Run ruff formatting and unit tests for all pull requests.

TeresiaOlsson:
I suggest that we update the workflow to run ruff formatting and unit tests for all pull requests and not just when pushing to main so we can catch the errors as early as possible.

### DIFF
--- a/.github/workflows/ruff_formatting.yml
+++ b/.github/workflows/ruff_formatting.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches: [ main ]
   pull_request:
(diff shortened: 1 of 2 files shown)


gubaidulinvadim (approved):
I'm surprised it's set up only for main. Anyhow, I think we've shared how to set up a git hook to check ruff on commit.

TeresiaOlsson:
Yes, if you install with the dev dependencies you also get the pre installation hooks. Its' only if you forget to install those dependencies that you can commit without it.

PR #333, opened 2026-07-13
Remove config models from lattice subpackage.

### LINKED ISSUE #319: Remove ConfigModel for lattice
This if for the lattice subpackage.

To do:

TeresiaOlsson:
The perhaps not so nice change here is the constructor of the `PyAtAttributeElementsLinker` where an object of `PyAtAttributeConfigModel` is needed to be created inside for the factory to be able to build the object.

This is to maintain the interface of `LatticeElementsLinker` which if I understand correctly has been written in a generic way to be able to expand with other linkers in the future. It therefore takes a generic `LinkerConfigModel` as input which can have an arbitrary number of attributes.

To make it nicer I can think of two options:

1. Remove the `LinkerConfigModel` and instead use `**kwargs` in `LatticeElementsLinker` to accept different number of configuration attributes.

2. Keep a configuration model and add an extra level in the configuration to specify the type configuration model to build.

Personally I feel that perhaps option 1 is the best to go for at this point since neither `LinkerConfigModel` or `LinkerIdentifier` define any requirements making them feel a bit obsolete.

### DIFF
--- a/pyaml/lattice/attribute_linker.py
+++ b/pyaml/lattice/attribute_linker.py
@@ -9,6 +9,8 @@ from pyaml.lattice.lattice_elements_linker import (
     LinkerIdentifier,
 )
 
+from ..validation import DynamicValidation, register_schema
+
 PYAMLCLASS = "PyAtAttributeElementsLinker"
 
 
@@ -45,7 +47,8 @@ class PyAtAttributeIdentifier(LinkerIdentifier):
         return f"{self.attribute_name}={self.identifier}"
 
 
-class PyAtAttributeElementsLinker(LatticeElementsLinker):
+@register_schema
+class PyAtAttributeElementsLinker(LatticeElementsLinker, DynamicValidation):
     """Link lattice elements using a specified PyAT element attribute.
 
     This linker associates PyAML elements with PyAT elements by comparing
--- a/pyaml/lattice/attribute_linker.py
+++ b/pyaml/lattice/attribute_linker.py
@@ -1,5 +1,6 @@
+from dataclasses import dataclass
+
 import at
-from pydantic import ConfigDict
 
 from pyaml.common.element import Element
 from pyaml.lattice.lattice_elements_linker import (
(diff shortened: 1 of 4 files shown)


TeresiaOlsson:
@gubaidulinvadim I rebased based on main to fix the merge conflicts but now merging is blocked because the integration tests are failing. But the errors look to me to have nothing to do with this PR and should work if it works on main. It is complaining about config files it can't find. Do you understand why it fails?

gubaidulinvadim:
> @gubaidulinvadim I rebased based on main to fix the merge conflicts but now merging is blocked because the integration tests are failing. But the errors look to me to have nothing to do with this PR and should work if it works on main. It is complaining about config files it can't find. Do you understand why it fails?

Yes, somehow we missed it on Friday. The old examples were deleted and the integration test fails because it was checking that examples run without errors

I'll open PR to fix it.

TeresiaOlsson:
> > @gubaidulinvadim I rebased based on main to fix the merge conflicts but now merging is blocked because the integration tests are failing. But the errors look to me to have nothing to do with this PR and should work if it works on main. It is complaining about config files it can't find. Do you understand why it fails?
> 
> Yes, somehow we missed it on Friday. The old examples were deleted and the integration test fails because it was checking that examples run without errors
> 
> I'll open PR to fix it.

But it still was merged into main? Maybe we then have something missing in our workflows? I thought something like that would block from merging into main?

But I also have another weird thing with the unit tests. For some reason they pass in the github workflow but not when I run it locally. Then `test_bba` fails despite that I'm according to git is running the same branch... Strange. So in the meantime I'm investigating why that happens. Maybe it could be because I locally do Python 3.11 and the workflow seems to run Python 3.12. If that's the case the workflow should maybe be modified so we test more python version. I think we decided to minimum support 3.11? But on PyPI we seem to claim that we support 3.9 and 3.10 too... Probably we need some PyPI cleanup too.

gubaidulinvadim:
> > > @gubaidulinvadim I rebased based on main to fix the merge conflicts but now merging is blocked because the integration tests are failing. But the errors look to me to have nothing to do with this PR and should work if it works on main. It is complaining about config files it can't find. Do you understand why it fails?
> > 
> > 
> > Yes, somehow we missed it on Friday. The old examples were deleted and the integration test fails because it was checking that examples run without errors
> > I'll open PR to fix it.
> 
> But it still was merged into main? Maybe we then have something missing in our workflows? I thought something like that would block from merging into main?

It was human error. We didn't pay attention because the PR was simply removing the files. Also, digital-twin tests are currently skipped locally and only checked on GitHub. That makes them a bit harder to check. Fixed in https://github.com/python-accelerator-middle-layer/pyaml/pull/356

Edit: but maybe something is missing. I think sometimes the digital twin tests are skipped also on GitHub. 

> But I also have another weird thing with the unit tests. For some reason they pass in the github workflow but not when I run it locally. Then `test_bba` fails despite that I'm according to git is running the same branch... Strange. So in the meantime I'm investigating why that happens. Maybe it could be because I locally do Python 3.11 and the workflow seems to run Python 3.12. If that's the case the workflow should maybe be modified so we test more python version. I think we decided to minimum support 3.11? But on PyPI we seem to claim that we support 3.9 and 3.10 too... Probably we need some PyPI cleanup too.

Open an issue

TeresiaOlsson:
Yes, it sounds like for some reason the digital twin test is not always running on github. Otherwise I think it wouldn't have been possible. I also only rely on github to run those tests for me. Maybe I do some detective work digging in the old actions later to understand why it happened.

Anyway, this PR is passing now after the fix :)

TeresiaOlsson:
I figured out the problem with the unit tests locally. No problem with the python version. Only that changes had been made to pySC and I had a too old version in my development environment since that doesn't automatically update.

gubaidulinvadim:
@TeresiaOlsson I left this PR gor Guillaume or Jean-Luc to review. I cannot quickly answer you questions in the top on two options for future implementation.

gupichon:
Hi everyone,

This part will need to be refactored anyway. It was an initial draft to allow using attributes other than FamName (for instance, we use UUIDs at SOLEIL).

The PR looks good to me, though we still need to better define the link between lattice and pyaml objects, as mentioned in discussion [#58](https://github.com/orgs/python-accelerator-middle-layer/discussions/58) by @gubaidulinvadim. In my opinion, this is a major topic.

gupichon (approved):

PR #334, opened 2026-07-13
Changes to remove config models from chromaticity.

### LINKED ISSUE #321: Remove ConfigModel for chromaticity
This is for:
- chromaticity_monitor
- chromaticity_response_matrix
- chromaticity

To do:

TeresiaOlsson:


gubaidulinvadim (approved):
_after_attach() for measurement tools and tuning tools seems less elegant than the previous solution. After attach would require to write it for each measurement tool and before I think this was not necessary.

PR #336, opened 2026-07-16
Changes for schema registry for magnets

### LINKED ISSUE #320: Remove ConfigModel for magnet
This if for the magnet subpackage.

To do:

If it turns out that one PR becomes too large it might need to be split.

TeresiaOlsson:


TeresiaOlsson:
This PR is almost completed. I just need to figure out the best way to keep the information about the inheritance when dynamically building ConfigurationSchemas since it turns out that virtual subclasses wasn't supported by pydantic... I'm adding a light-weight version of the same now since it is just needed for the validation and JSON schema generation which anyway are both custom for us.

gubaidulinvadim (approved):

Issue #341, opened 2026-07-24
Remove ConfigModel for BBA


TeresiaOlsson:
The BBA is a new module which hasn't been included in the changes.

This also needs to be updated after #335 has been merged in.

PR #335, opened 2026-07-14
Changes for schema registry for tuning tools

TeresiaOlsson:
I decided to make the modifications of the tuning tools in one PR because the response matrices, measurement tools and tuning tools affect at many places.

Edit from @gubaidulinvadim : Resolves #321 #322 #323 #324 #325 #326

PR #346, opened 2026-08-12
Updated BBA2 algo

JeanLucPons:
This PR switches to a more standard BBA algorithm (still model dependent).
With optional optimum checking.

Below a part of the review done by Claude AI. This was discussed with @swhite2401 and this is the behavior we want.
The idea is to let more flexibility to the user to redo a scan after eventually applying offset and correcting the orbit in order to reduce systematic error. It is easy to the user to check if the optimum steerer value is inside or outside [-dk0h, dk0h].

> Design tradeoff, not a bug: the new fixed sweep assumes the true offset lies within the configured [-dk0h, dk0h] range, whereas the old adaptive algorithm could walk outward if the initial guess was off. That's a reasonable simplification for a "more standard" algorithm, but it does put more weight on hcorr_delta/vcorr_delta being well-chosen — might be worth a docstring note or a sanity check/warning if the fit result falls outside the measured step range (the old code had a don't rely on extrapolation comment/check that doesn't appear to have an equivalent here anymore).

### DIFF
--- a/pyaml/tuning_tools/bba2.py
+++ b/pyaml/tuning_tools/bba2.py
@@ -76,7 +76,6 @@ class BBAData:
         self.bpm_pos = []  # BPM#i position
         self.allbpm_pos = []  # IOS (Induced Orbit Shift)
         self.steps = []  # steerer step
-        self.fit = []  # linear fit
 
         self.opt_type = BBAData.OPT_TYPE.NONE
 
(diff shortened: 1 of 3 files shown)


gubaidulinvadim:
@JeanLucPons https://github.com/python-accelerator-middle-layer/pyaml/pull/335 was merged into main and it removes ConfigModel from the tuning tools. This might affect your PRs on BBA.

TeresiaOlsson:
> @JeanLucPons #335 was merged into main and it removes ConfigModel from the tuning tools. This might affect your PRs on BBA.

It will but I will make the PR for #341 now and then I think it should be easy to rebase this PR based on that.

Issue #199, opened 2026-02-20
ElementHolder API refurbishment


JeanLucPons:
**Description, motivation and use case**

Today we have a single `ElementHolder` class that holds everything for a given control mode.
The `ElementHolder` suffers from a large number of methods. There is a need for reorganization and the introduction of other holders. Last but not least, `ElementHolder.get_all*()` or `ElementHolder.get_*s()` function return list while typed array would be suitable. The  `ElementHolder` will however hold all elements as before.
Note that the `get` function from typed holder will return appropriate typed object or arrays and raise an exception in case of wrong type while the equivalent `get` on the base ElementHolder may return random types.

```python
class MagnetsHolder(Object):
  # Return the named array or all magnets when no name specified
  # sr.magnets[:] is equivalent to sr.magnets.get() 
  def get(name:str = None) : "MagnetArray"
     ...

class MagnetHolder(Object):
  # Return the magnet with the specified name
  def get(name:str) : "Magnet"
     ...
 ```

**Proposed solution**

Create new holders such as `Magnet(s)Holder`, `BPM(s)Holder`, `RFHolder`, `DiagnosticHolder`,`ToolHolder` etc and allow access to them trough python properties to allow following writings:
This is a non exhaustive list.

```python
#Magnet(s)Holder
sr.live.magnets.get() # Return all magnets simple and virtual (no combined function magnet)
sr.live.magnets[:] # Return all magnets simple and virtual (no combined function magnet)
sr.live.magnets.get_cfm() # Return all combined function magnets
sr.live.magnets.get("QuadForTune") # Returns family `MagnetArray`
sr.live.mangets.QuadForTune  # If we add dynamic python properties
sr.live.magnets['*QD*'] # Returns subset
sr.live.magnets.get("QuadForTune")['QD1*'] # Returns subset of familly
sr.live.magnets[1:10] # Returns subset

sr.live.magnets.get()[model_name:'*-H'] # Returns subset
sr.live.manget.get("QF1E-C03").strength.set( 0.8 ) # Exmaple with single manget

# BPM(s)Holder (similar to MagnetHolder)
sr.live.bpms.get(...)
sr.live.bpm.get(...)

# RFHolder
sr.live.rf.masterclock.frequency.set()  # The `RFHolder` will contain a masterclock object coming from the "DEFAULT_RF_PLANT"
sr.live.rf.get(rfname) # Will return an other RFPlant

# ElementHolder

[...]

PR #347, opened 2026-08-14
New magnet holder api

JeanLucPons:
This PR implements new holder API for magnet(s).
Part of [#199](https://github.com/python-accelerator-middle-layer/pyaml/issues/199)


pyaml/common/magnets_holder.py
```diff
@@ -0,0 +1,39 @@
...
+
+if TYPE_CHECKING:
+    from .element_holder import ElementHolder
+
+
+class MagnetsHolder:
+    def __init__(self, peer: "ElementHolder"):
+        self._peer = peer
+
+    def get(self, name: str = None) -> Magnet:
```
gubaidulinvadim: Should return an array type?

pyaml/common/magnets_holder.py
```diff
@@ -0,0 +1,39 @@
...
+
+        Parameters
+        ----------
+        name : str
+            Name of the magnet array
+        """
+        if name is None:
+            return [value for key, value in self._peer._MAGNETS.items()]
+        else:
+            return self._peer._get("Magnet array", name, self._peer._MAGNET_ARRAYS)
```
gubaidulinvadim, line 25: I'm not sure if this should be part of this PR, but we would also like to have a functionality to get a list of magnets that match a regex/wildcard string. At the moment, it's only possible to define a wildcard or regex in the configuration file, but not to search by code.

```
element_holder.magnets.get('AE/C01*QUAD*')
```

gubaidulinvadim (dismissed):
Looks good. The changes are actually minimal to make a subholder.

JeanLucPons:
I added the selection as specified in #199 
```
arr = sr.design.magnets["QD2E-C3*"]
```
Selection by other field such as `device name` will required that we first agreed on `__repr__`. For the time being __repr__ are not coherent between various objects.

gubaidulinvadim (approved):

PR #358, opened 2026-08-24
Update workflow for unit tests to test against several python versions

TeresiaOlsson:
I have updated the github workflow to run the unit tests for python 3.11-3.14 to automatically be able to catch any compatibility issues.

I also upgraded the github actions to use the latest versions.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -7,6 +7,7 @@ on:
   push:
     branches: [ "main" ]
   pull_request:
+  workflow_dispatch:
 
 permissions:
   contents: read
@@ -18,7 +19,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.11", "3.12", "3.13", "3.14"]
+        python-version: >-
+          ${{ (github.event_name == 'push' || github.event_name == 'workflow_dispatch')
+              && fromJSON(vars.PYTHON_ALL_SUPPORTED_TEST_VERSIONS)
+              || fromJSON(vars.PYTHON_DEFAULT_TEST_VERSIONS) }}
 
     steps:
     - uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


TeresiaOlsson:
@gubaidulinvadim I noticed that the twin tests don't run for this PR. ChatGPT claims it's because:

```
    paths:
      - ".github/workflows/dt4acc-integration.yml"
      - "pyaml/**"
      - "tests/integration/**"
      - "pyproject.toml"
```

which only makes this workflow run if there are changes to those specific paths. So that likely explains what happened when the examples where changed.

gubaidulinvadim:
I think it could be good to have it as an optional on-demand workflow. Triggering it on each pull request may not be necessary.

TeresiaOlsson:
> I think it could be good to have it as an optional on-demand workflow. Triggering it on each pull request may not be necessary.

You mean for the python versions or for the twin tests? For the python versions we could probably do so it's just run for one version for a PR and only for all versions when doing the merge into main.

gubaidulinvadim:
> > I think it could be good to have it as an optional on-demand workflow. Triggering it on each pull request may not be necessary.
> 
> You mean for the python versions or for the twin tests? For the python versions we could probably do so it's just run for one version for a PR and only for all versions when doing the merge into main.

I mean unit tests for Python versions. I think this is only necessary when merging into main. On GitLab there's a way to do optional "action", I think GitHub can do the same. I was thinking to have Python 3.12 (or other) as default and others could be triggered by user.

TeresiaOlsson:
Makes sense. I will add that. But maybe the default to always run for should be 3.11 since that's the minimum version we said to support?

TeresiaOlsson:
I modified so now it should only run the versions defined by PYTHON_DEFAULT_TEST_VERSIONS during PR and all defined by PYTHON_ALL_SUPPORTED_TEST_VERSIONS during merge to main and if manually triggered in the web interface. I created both those variables at organizational level so they can also be used by other repositories.

The versions can be change here: https://github.com/organizations/python-accelerator-middle-layer/settings/variables/actions instead of modifying the workflow when we in the future want to change which versions to support.

gubaidulinvadim (approved):

PR #367, opened 2026-09-04
Update docstrings using codex

TeresiaOlsson:
I told codex to update the docstrings everywhere so they are numpy style and exists on module, class and method level.

It is in no way perfect and I haven't proof-read all the details. Just gone through it so nothing very strange was added and iterated the parts was generated to generic the first time.

I hope this will provide a starting basis that we can use to improve it over time.


simoneliuzzo:
Dear @TeresiaOlsson,

that is a nice work! Where there was nothing now there is something and in the right format! 

I see plenty of "Input value for this operation." But this is for us to fix with the relevant text.

TeresiaOlsson:
> Dear @TeresiaOlsson,
> 
> that is a nice work! Where there was nothing now there is something and in the right format!
> 
> I see plenty of "Input value for this operation." But this is for us to fix with the relevant text.

Maybe it can also be fixed by running codex some more iterations on the parts that are not so good. I already did it with some sections and found that it produced better results when it was given a smaller section.

I have rebased based on main to fix the merge conflicts and I also had to delete the github workflow for building the documentation since readthedocs now automatically understands to do it.

Can you approve it now so it can be merged in?

GamelinAl:
Hi @TeresiaOlsson , I am trying to do a second pass to fix the different issues I see.
I'll commit it soon so you can have a look before the merge

GamelinAl:
@TeresiaOlsson I made a 1st pass but there are still a few things that I would like to change before this is merged.
But I would like to know if you agree:

1. Add method section to Class docstring (only the public methods, not the ones starting from _ or __)
2. Move parameter section to Class docstring instead of __init__ (and remove duplication)

The users will run `help(Chromaticity)` not `help(Chromaticity.__init__)`. Most IDE and IPython help render the class docstring so this is here the information should be.

TeresiaOlsson:
> @TeresiaOlsson I made a 1st pass but there are still a few things that I would like to change before this is merged.
> But I would like to know if you agree:
> 
> 1. Add method section to Class docstring (only the public methods, not the ones starting from _ or __)
> 2. Move parameter section to Class docstring instead of __init__ (and remove duplication)
> 
> The users will run `help(Chromaticity)` not `help(Chromaticity.__init__)`. Most IDE and IPython help render the class docstring so this is here the information should be.
> 
> 

Yes, that sounds good. Even better would be if we somehow could put everything we want about the format in a file that we save somewhere (maybe in the copier template?). Then we can easily tell codex to read the file and produce the same format for all the other packages and we have similar starting points for docstrings everywhere.

GamelinAl:
@TeresiaOlsson please check that the changes are ok with you.

GamelinAl (approved):

TeresiaOlsson:
I have merged this and the PR for the new api documentation now. As soon as the readthedocs build has finished we should be able to see when when choosing the latest version what the current state of the API documentation is in main now. Feels likely that we will find more things that can be improved.

https://pyaml.readthedocs.io/en/latest/api.html

Issue #382, opened 2026-09-08
ResponseMatrixData "type" field overwrites "type" to None.


kparasch:
@TeresiaOlsson Is this field in ResponseMatrixData required? (Unfortunately I am not yet familiar with how DynamicValidation works)
Because of this field, when saving a response matrix the type gets overwritten to None.

https://github.com/python-accelerator-middle-layer/pyaml/blob/f8b1a2337759d60970d3f46211980183ff00126a/pyaml/tuning_tools/response_matrix_data.py#L34

E.g. when running `pyaml/examples/use_cases/03-orbit_correction.py`. I checked that the response matrix is saved and reloaded correctly when removing this field definition, but I am not sure if something else may be breaking.

TeresiaOlsson:
The dynamic validation just creates a model to validate against from the constructor of the class. It should save what it created in an attribute `validation_model` if needed to check what it generated.

Should there be a field `type` in the data? It looked to me from the old example data which existed that there can be? If so, the field should be there otherwise the data model is not complete. The validation should then fail if you try to create a new object of the class where there is `type` in the data. But not if  `type` is missing because it accepts `None` for it. Is that the behaviour you see?

Strange though why this would affect the save. Maybe there is a bug in the save function which always put the type to `None`?

kparasch:
The `type` was there to find the model with which to do the validation.

Just like in 
```
type: pyaml.accelerator
machine: sr
facility: BESSY2
energy: 1.7185e9
simulators:
- type: pyaml.lattice.simulator
  lattice: bessy2_standard_user.mat
  name: design
.
.
.
```

It is not really part of the actual data.

TeresiaOlsson:
So it should be removed both from the data model and the existing files?

When I saw it I actually quite liked that the field was there because it reminded me of MML where it in the data sometimes saves information about which function created it. So if you have a file with a random data you can look at that field and understand what type of data it is. But then the field becomes part of the data and should be mandatory so the option to have it as `None` should be removed.

kparasch:
I think I am outdated on how the factory works.

It's not part of the data model, but it should still be written in the existing files.

According to factory.py, I need to define `type` or `class` in the configuration file. So even though "type" is not part of the orbit response matrix data, it should still be there in the output file so that the factory is able to build it.

from factory.py:
```
        # Legacy module/class format:
        #   type: some.module
        #   class: MyClass        (optional; otherwise PYAMLCLASS is used)

[...]

PR #383, opened 2026-09-08
Bugfix/orm saving

kparasch:
ResponseMatrixData field was triggering a bug where it was setting type to None when saving ORM data.

Extended the test to catch the error, and removed ResponseMatrixData.type field.

This solves #382

### DIFF
--- a/pyaml/tuning_tools/response_matrix_data.py
+++ b/pyaml/tuning_tools/response_matrix_data.py
@@ -2,6 +2,7 @@ from dataclasses import dataclass
 from pathlib import Path
 
 from pyaml.configuration.factory import Factory
+from pyaml.validation import SchemaValidator
 
 from .. import PyAMLException
 from ..configuration.fileloader import load
@@ -54,6 +55,7 @@ class ResponseMatrixData(DynamicValidation):
         path = Path(filename)
         if path.exists():
             config_dict = load(str(path.resolve()))
+            SchemaValidator.validate(config_dict)
             return Factory.build(config_dict, ignore_external=False)
         else:
             raise PyAMLException(f"{filename}: file not found")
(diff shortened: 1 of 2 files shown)


TeresiaOlsson:
Does the change mean that the `type` still need to exist in the data file or not?

If it still needs to exist, I think there is another problem because there is currently nothing which will verify that it does. The validation of if there is a `type/class` field happens when loading the configuration but it will not check fields inside files that aren't loaded yet.

If so I think there are two options:

1. Change to not use the factory to build the object.

2. Add a validation step in `load` which validates the data after the file content is loaded against the ConfigurationSchema for ResponseMatrixData. That model includes the  `type/class` field so the validation will check that is exists before passing the data to the factory.

kparasch:
The factory will check if there is "type" or "class". If it doesn't find it, I understand it will throw an exception. I am still trying to understand how to trigger the validation.

JeanLucPons (dismissed):

kparasch:
Don't merge, I just found another bug.

TeresiaOlsson:
> The factory will check if there is "type" or "class". If it doesn't find it, I understand it will throw an exception. I am still trying to understand how to trigger the validation.

Maybe I can use you as the tester for this documentation page ;) https://python-accelerator-middle-layer.github.io/documentation/how-to/configuration/use-schema-registry.html

It should explain how to use the new validation functionality. I think this is what you need: https://python-accelerator-middle-layer.github.io/documentation/how-to/configuration/use-schema-registry.html#validate-configuration

kparasch:
Thanks! I found the documentation (I only checked this part) quite nice, and I actually liked SchemaRegistry and SchemaValidator! 

I was actually getting annoyed with jumping all the time through all the files to check what fields are necessary for which class. SchemaRegistry gives a nice directory.

P.S. False alarm on the bug I mentioned before.

JeanLucPons (approved):

gubaidulinvadim:
> ResponseMatrixData field was triggering a bug where it was setting type to None when saving ORM data.
> 
> Extended the test to catch the error, and removed ResponseMatrixData.type field.
> 
> This solves #382

It was set to None to avoid some type-check or validation error after the configuration was changed. There was an error for the type between SC and pyAML. I thought that it's pointless to have a type check for SC responsematrix. So, I did a lazy fix of setting it to None somewhere.

PR #385, opened 2026-09-08
Bugfix: array type inference

### LINKED ISSUE #384: Array type inference issue for mixed magnet classes
4. magnets[:] and magnets["Q*D*"] works but return ElementArray instead of MagnetArray

```
type(sr.design.magnets[":"])
>>> pyaml.arrays.element_array.ElementArray
```

5. Similar bug with using subset of magnet arrays
```
magnets.get()        -> MagnetArray     ✅
magnets.get()[:]     -> ElementArray    ❌ should be MagnetArray
magnets.get()['S*']  -> ElementArray    ❌ should be MagnetArray
```

_Originally posted by @GamelinAl in https://github.com/python-accelerator-middle-layer/pyaml/issues/376#issuecomment-5582439953_

gupichon:


GamelinAl (approved):
I checked on the examples which were failing before i.e. #384 now it works for me.

PR #389, opened 2026-09-09
Report real constructor signatures on `ValidationMeta` classes

### LINKED ISSUE #387: ValidationMeta.__call__ masks constructor signatures
While looking at the new documentation I found something strange with some classes which online documentation did no reflect its docstring.

Here is a quick AI summary of the issue.

## Summary

`ValidationMeta.__call__(cls, *args, **kwargs)` (`pyaml/validation/validation_models.py:41`)
replaces the effective call signature of every class built on it. Because
`inspect.signature(SomeClass)` follows the metaclass `__call__` before falling back to
`__init__`, all classes deriving from `DynamicValidation` / `StaticValidation` introspect
as `(*args, **kwargs)`:

```python
>>> import inspect
>>> from pyaml.magnet.quadrupole import Quadrupole
>>> inspect.signature(Quadrupole)
(*args, **kwargs)          # actual: (name, model=None, lattice_names=None, description=None)
>>> inspect.signature(Quadrupole.__init__)
(self, name: str, model: MagnetModel | None = None, ...)   # the real one, still intact
```

46 of the 173 public classes are affected (every module importing
`DynamicValidation`/`StaticValidation`).

## Impact

1. **Rendered docs are wrong.** `sphinx.ext.autodoc` uses the same introspection, so the
   HTML API pages show `class Quadrupole(*args, **kwargs)` regardless of how complete the
   docstring is. The Parameters section is documented, but the signature line above it is
   useless, and `autodoc_typehints = "description"` has nothing to pull types from.
2. **False numpydoc validation errors.** The remaining 19 `PR02` (unknown parameters) and
   27 `PR01` (undocumented parameters) reported against these classes are artifacts of the
   masked signature — numpydoc compares the docstring's Parameters against `(*args, **kwargs)`.
   They are not missing docs, and they cannot be fixed by editing docstrings.
3. Anything else relying on `inspect.signature` of the class (IDE completion, help(),
   downstream tooling, schema generation from the class rather than `__init__`) sees the
   same placeholder.

## Proposed fix

Give `ValidationMeta` a `__signature__` property that reports the signature of
`cls.__init__` with `self` dropped, e.g.:

```python
class ValidationMeta(ABCMeta):
    @property
    def __signature__(cls):
        sig = inspect.signature(cls.__init__)
        return sig.replace(parameters=list(sig.parameters.values())[1:])
```

Points to settle during implementation:

- Whether to advertise the `validate: bool = True` keyword that `__call__` pops
  (`validation_models.py:68`) as a trailing keyword-only parameter, so the documented
  signature matches what the class actually accepts. Probably yes, but it then appears on
  all 46 classes.
- Caching: `inspect.signature` on each access is cheap but not free; a `functools.cache`d
  helper keyed on the class may be worth it if this is hit in hot paths.
- `__init__` resolution should go through the MRO (classes such as `Quadrupole` define
  their own, but not all subclasses do) — `inspect.signature` already handles that.

GamelinAl:
Closes #387
Code fully generated by AI, I checked that the fix seem to be working but I don't really know the specifics of this part of the code.
The main choice in the fix is **not** to expose the `validate` keyword in all signature of the concerned class.

## Problem

`inspect.signature()` consults a class's metaclass `__call__` before falling back to
`__init__`. Since `ValidationMeta.__call__(cls, *args, **kwargs)` fronts every class built on
`DynamicValidation` / `StaticValidation`, **46 of the 173 pyaml classes** introspected as
`(*args, **kwargs)`:

```python
>>> inspect.signature(Quadrupole)
(*args: Any, **kwargs: Any)
```

## Change

A `__signature__` property on `ValidationMeta` returning the signature of `cls.__init__` 
without `self`. As a property on the metaclass it is a data descriptor on `type(cls)`, so it
wins the `Quadrupole.__signature__` lookup that `inspect.signature` performs before its
metaclass-`__call__` branch. Details:

- `inspect.signature(cls.__init__)` already walks the MRO, so subclasses without their own
  `__init__` get the inherited one, and the `@dataclass` classes resolve correctly.
- Classes relying on `object.__init__` report an empty signature, which is what `inspect`
  reports for a plain class.
- The `validate` keyword that `__call__` pops is deliberately **not** advertised: including it
  would make numpydoc flag `validate` as undocumented on all 46 classes. It remains documented
  in the `ValidationMeta` and `DynamicValidation` docstrings, and the `ValidationMeta` docstring
  now states that the reported signature is `__init__`'s.
- Failure to compute a signature returns `None`, which `inspect` treats as "absent", falling
  back to the previous behaviour rather than raising.
- No caching: no internal code path calls `inspect.signature(cls)`. `configuration/manager.py`,
  `validation/schema_builder.py` and `ValidationMeta.__call__` itself all go through
  `cls.__init__` explicitly and are unaffected.

No behavioural change: validation, positional/keyword binding and `validate=False` are untouched.

## Tests

Seven tests in `tests/validation/test_models.py` covering the dynamic and static branches,
inherited constructors, `@dataclass` subclasses, classes without an `__init__`, exclusion of
`validate`, and non-leakage of `__signature__` onto instances.

## Verification

| Check | Result |
|---|---|
| Sweep of all 173 pyaml classes | 0 of the 46 still report `(*args, **kwargs)` |
| `pytest tests` | 279 passed, 7 skipped |
| numpydoc over the 46 classes | `PR01` 46 → 2, `PR02` 44 → 0 |
| Sphinx HTML | `class …Quadrupole(name, model=None, lattice_names=None, description=None)` |
| `ruff format --check` / `ruff check` | clean |

The 2 remaining `PR01` are genuine: `ResponseMatrixData` and `OrbitResponseMatrixData` do not
document their `type` parameter. They are left for a separate docstring change.

### DIFF
--- a/pyaml/validation/validation_models.py
+++ b/pyaml/validation/validation_models.py
@@ -36,7 +36,41 @@ class ValidationMeta(ABCMeta):
     invoked.
 
     Pass `validate=False` to skip validation for a single construction.
+
+    The signature reported by :func:`inspect.signature` for classes using this
+    metaclass is that of their ``__init__``, not of :meth:`__call__`. The
+    ``validate`` keyword is therefore not part of the reported signature.
     """
+
+    @property
+    def __signature__(cls) -> inspect.Signature | None:
+        """
+        Report the constructor signature of the class.
+
+        Without this, :func:`inspect.signature` resolves to :meth:`__call__` and
+        reports ``(*args, **kwargs)`` for every class using this metaclass, which
+        hides the real parameters from ``help()``, Sphinx and other tooling.
+
+        Returns
+        -------
+        inspect.Signature | None
+            Signature of ``cls.__init__`` without ``self``, or ``None`` if it
+            cannot be determined, in which case the default introspection applies.
+        """
+        # Classes relying on ``object.__init__`` take no argument, as reported by
+        # ``inspect.signature`` for a class without a metaclass
+        if cls.__init__ is object.__init__ and cls.__new__ is object.__new__:
+            return inspect.Signature()
+
+        try:
+            signature = inspect.signature(cls.__init__)
+        except (TypeError, ValueError):  # pragma: no cover - defensive
+            return None
+
+        # Drop ``self`` and the constructor return annotation
+        parameters = list(signature.parameters.values())[1:]
+
+        return signature.replace(parameters=parameters, return_annotation=inspect.Signature.empty)
 
     def __call__(cls, *args: Any, **kwargs: Any):
         """
--- a/tests/validation/test_models.py
+++ b/tests/validation/test_models.py
@@ -1,6 +1,8 @@
 """Tests of the configuration models."""
 
+import inspect
 import json
+from dataclasses import dataclass
 
 import pytest
 from pydantic import BaseModel, ValidationError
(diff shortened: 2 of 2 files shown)


TeresiaOlsson:
I can review this later today. I just want to first fix all the bugs related to the validation that are appearing now because we merged in the changes to remove the config models from pyaml-cs-oa + adding dynamic validation on the accelerator.

Currently the integration tests for this PR is also failing because of that.

TeresiaOlsson:
The integration tests are passing again now :)

TeresiaOlsson (approved):

PR #391, opened 2026-09-09
Change type to class in configuration files

TeresiaOlsson:
I have made the updates so the default key in the configuration is class_path/class instead of type. Type is still supported so configuration files which are using the old format still is possible to load.

I could only change the BESSY II example for now since the others need PR https://github.com/python-accelerator-middle-layer/tango-pyaml/pull/35 to be merged before they can be validated.

I also didn't change the configurations in the tests for now to avoid breaking it before the hackathon. For now I mostly want to update the files that users might look at to learn how to write the configuration so they learn the new format.

### DIFF
--- a/examples/use_cases/config/bessy2/bessy2.yaml
+++ b/examples/use_cases/config/bessy2/bessy2.yaml
@@ -1,21 +1,21 @@
-type: pyaml.accelerator
+class: pyaml.accelerator.Accelerator
 machine: sr
 facility: BESSY2
 energy: 1.7185e9
 simulators:
-- type: pyaml.lattice.simulator
+- class: pyaml.lattice.simulator.Simulator
   lattice: bessy2_standard_user.mat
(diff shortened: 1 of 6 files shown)


gupichon:
Could you merge main into your branch to make sure everything is consistent with the recent changes?

TeresiaOlsson:
> Could you merge main into your branch to make sure everything is consistent with the recent changes?

I rebased it on the current main now.

JeanLucPons:
> > Could you merge main into your branch to make sure everything is consistent with the recent changes?
> 
> I rebased it on the current main now.
Thanks

@gupichon 
rebase + force push is the good way to do in a PR otherwise it blends history in the main and in your branch and some old code can reappear in conflict resolutions. I already experienced such issue. So please , in a PR, use rebase.

JeanLucPons (approved):

PR #397, opened 2026-09-10
Add workflow for github release.

TeresiaOlsson:
I have added the workflow for making the github release. It's the same as in the other packages.


TeresiaOlsson:
Turns out there was already a PR for this...

PR #404, opened 2026-09-11
Fix new holder API in tune example, removed catalog which is the default

JeanLucPons:
This PR fix new holder API (get_magnets) in tune exmaple and use default dynamic catalog.
Require: https://github.com/python-accelerator-middle-layer/pyaml-cs-oa/pull/49


JeanLucPons:
@simoneliuzzo 
You can see the config change in the diff here

simoneliuzzo (approved):

PR #407, opened 2026-09-11
Modified dependencies because pypi didn't accept direct dependencies.

TeresiaOlsson:
I had to make a hotfix to the release because the action to publish on pypi failed. This commit has already been applied to stable and is in the new release. This PR is to put it into main too.


TeresiaOlsson:
@gupichon Looks like the integration tests are failing now but I don't understand why. Do you understand it?

gupichon:
The test will pass if you merge from main.

TeresiaOlsson:
Fixed by #424

PR #417, opened 2026-09-14
Track latest PyPI releases of tango-pyaml/pyaml-cs-oa in dt4acc CI

### LINKED ISSUE #416: BuxFix: dt4acc-integration.yml pins tango-pyaml/pyaml-cs-oa to fixed git refs instead of tracking the latest PyPI release
## Context

.github/workflows/dt4acc-integration.yml, step "Install pyAML and control-system backends", installs the two companion control-system packages from fixed git refs instead of PyPI:

python -m pip install --no-deps \
  "tango-pyaml @ git+https://github.com/python-accelerator-middle-layer/tango-pyaml.git@1f010735bd675ad2f918958abd0396cb6766f14e" \
  "pyaml-cs-oa @ git+https://github.com/python-accelerator-middle-layer/pyaml-cs-oa.git@0.2.0"

- tango-pyaml is stuck on a specific commit hash.
- pyaml-cs-oa is stuck on git tag 0.2.0, fetched via git+https://... rather than PyPI.

## Problem

These pins go stale: every time a new release of tango-pyaml or pyaml-cs-oa is published, the dt4acc integration pipeline keeps testing against the old, frozen ref until someone remembers to manually bump the commit/tag in the workflow. This defeats the purpose of the smoke test, which is meant to catch incompatibilities between pyaml and the current published companion packages — not a snapshot from whenever the pin was last touched.

## Expected behavior

The workflow always installs the latest published PyPI release of tango-pyaml and pyaml-cs-oa, with no fixed version/commit pin, so the integration test naturally stays current as new releases come out.

## Proposal

Replace the --no-deps git install with an unpinned PyPI install:

python -m pip install --no-deps tango-pyaml pyaml-cs-oa

(keep --no-deps if it's intentional here to avoid pulling unwanted transitive deps into this job; otherwise reconsider it too).

## Acceptance criteria

- [ ] dt4acc-integration.yml installs tango-pyaml and pyaml-cs-oa from PyPI, unpinned (always latest), with no git+https://... source and no version pin.
- [ ] The dt4acc smoke test still passes against the currently published PyPI versions.
- [ ] (Nice to have) if the smoke test starts failing after a new companion release, that failure is attributable to the new release — e.g. log the installed version(s) in the workflow output for easier triage.

gupichon:
Install these companion packages unpinned from PyPI instead of fixed git refs (commit hash / tag), so the dt4acc smoke test stays current with newly published releases instead of testing a stale snapshot.

Fixes #416


JeanLucPons (dismissed):

gupichon:
@TeresiaOlsson, @JeanLucPons, what do you prefer? Should I point it to:

1. The latest release on PyPI, which means downgrading the BESSY II example to make it work.
2. The latest on GitHub, but this prevents users from having an example that actually works with the PyPI release (as is currently the case).

Personally, I lean toward option 1, but I’m happy to go with the consensus. I don't have a strong opinion either way.

gupichon:
@JeanLucPons, thanks for the review! Just a heads-up: I'll need to update the BESSY II example to match the PyPI version, which will dismiss your review.

JeanLucPons:
I agree with point 1

TeresiaOlsson:
> I agree with point 1

Me too. It feels more robust to me to always test against latest release. Then at least the tests runs against a stable source.

gubaidulinvadim (dismissed):

gupichon:
@gubaidulinvadim, the test is fine now.

gubaidulinvadim (approved):

PR #427, opened 2026-09-16
Add an RF masterclock facade

### LINKED ISSUE #374: Feature: Add an RF masterclock facade
`RFHolder` already resolves named RF plants and exposes frequency and voltage from `DEFAULT_RF_PLANT`. The original holder proposal expresses this default RF plant as `rf.masterclock`, which makes its role explicit and leaves room for other RF plants under the same holder.

This improves readability in control applications that operate on the machine master clock while still supporting named RF plants.

**Proposed solution**
Add `rf.masterclock`, returning the RF plant configured as `DEFAULT_RF_PLANT`:

- `rf.masterclock.frequency` exposes its frequency access.
- `rf.masterclock.voltage` exposes its total-voltage access.
- `rf.get(name)` continues to resolve other RF plants by name.

Retain `rf.frequency` and `rf.voltage` as backward-compatible aliases for the corresponding masterclock accessors. Raise a clear `PyAMLException` when `DEFAULT_RF_PLANT` is absent.

Keeping only `rf.frequency` and `rf.voltage` is compact but hides which configured RF plant they refer to. Making `masterclock` a separate object would duplicate the `RFPlant` API and create unnecessary synchronization concerns.

**Example**
```python
masterclock = sr.live.rf.masterclock
masterclock.frequency.set(499.654e6)
masterclock.voltage.set(2.5e6)

same_frequency = sr.live.rf.frequency
spare_rf_plant = sr.live.rf.get("SPARE_RF_PLANT")
```

**Additional context**
Parent issue: [ElementHolder API refurbishment #199](https://github.com/python-accelerator-middle-layer/pyaml/issues/199).

The current `RFHolder` already implements `get(name)`, `frequency`, and `voltage`. This issue introduces the explicit default-plant façade proposed in #199 while retaining the current shortcuts.

**Checklist**

gupichon:
## Description

Adds `RFHolder.masterclock`, returning the `RFPlant` configured as `DEFAULT_RF_PLANT`, making the default RF plant explicit. `frequency`/`voltage` become backward-compatible aliases for `masterclock.frequency`/`masterclock.voltage` instead of duplicating the `DEFAULT_RF_PLANT` lookup.

## Related Issue

- Fixes #374

- [x] new feature: `rf.masterclock` was added as a property delegating to the existing `get("DEFAULT_RF_PLANT")` — no new lookup/error logic needed, since `ElementHolder._get()` already raises `PyAMLException` when the name is absent (same convention already used for `chromaticity`/`tune`/`orbit`/etc.).

- [x] `RFHolder.frequency`/`RFHolder.voltage`: reimplemented to delegate to `self.masterclock.frequency`/`self.masterclock.voltage` instead of calling `self.get("DEFAULT_RF_PLANT")` directly, to avoid duplicating the literal name in two places. Behavior and return values are unchanged for existing callers.

## Testing

The following tests (compatible with pytest) were added to `tests/rf/test_rf.py`:
- [x] `test_masterclock_returns_the_default_rf_plant`
- [x] `test_frequency_and_voltage_are_masterclock_aliases`
- [x] `test_masterclock_raises_when_default_rf_plant_missing`
- [x] `test_simple_rf_access`

- [x] New and existing unit tests pass locally (308 passed, 1 pre-existing unrelated failure: `tests/tuning_tools/test_tuning_orm.py`, a Windows-only `NamedTemporaryFile` double-open issue)
- [x] The code is commented where appropriate (NumPy-style docstrings, including the class-level docstring so `masterclock` is discoverable via `help(sr.live.rf)`)


JeanLucPons:
To me this is a bit confusing, a masterclock does not control the RF voltage.

gubaidulinvadim:
> To me this is a bit confusing, a masterclock does not control the RF voltage.

@gupichon I agree with @JeanLucPons , I wrote the same thing in a linked issue a few days ago.

gubaidulinvadim:
@gupichon this PR is also strange because I see a green MERGE button without any reviews.
Edit: OK this is merged into unprotected branch, this is fine.

JeanLucPons:
Concerning the voltage (or rather the power) of the masterclock, it is tuned to ensure that the signal goes through all cables to be detected by all low level RF elements. Then the RF voltage in various cavity (booster or SR) is regulated by the RF transmitters that include their own low level RF element.

So the masterclock power tuning is rather for RF specialist and not for accelerator physicists. To me it should not be in pyAML.
We can even imagine to transmit the masterclock signal using numeric system such as white rabbit timing system.

GamelinAl:
To me the `masterclock` device should only contain the `frequency` attribute. I agree with @JeanLucPons  and @gubaidulinvadim that a `voltage` atribute for a `masterclock` does not make sense.

I think we should have a `cavity` device controlling the `voltage` and `phase` (corresponding to the total cavity voltage and cavity phase for the sum of all cavities for a given harmonic). If the synchrotron use only a single harmonic the `phase` attribute is probably not useful, but it is for multi-harmonic systems.

Both of these different elements could share the `RFHolder`.

JeanLucPons:
The actual RF holder provides:

```python
sr.live.rf.frequency.set(...)  # Set main RF freq (the one of the DEFAULT_RF_PLANT, for powered harmonic cav you should pass by its transmitter)
sr.live.rf.voltage.set(...) # Set the total RF voltage of all main RF freq (not including powered harmonic cav)
sr.live.rf.transmitter.get("TRA1").voltage.set(...) # Set the voltage of cavities powered by TRA1
sr.live.rf.transmitter.get("TRA1").phase.set(...) # Set the phase of cavities powered by TRA1
```

Edit:
sr.live.rf.frequency.set()  can also affect powered harmonic cavity, the harmonic order is already a part of the config but (as least for us) it is plan to do that by hardware. So i don't think we will need to tune the frequency of powered harmonic cav.

gupichon:
Thanks everyone for the feedback. If I understand correctly, I should just revert voltage back to how it was before this PR, and we'll design a separate cavity feature later to cover voltage/phase. Is that right?

gubaidulinvadim:
@gupichon I think this PR could just be closed. Presently the rf.frequency.get() should already be the masterclock frequency.

JeanLucPons:
I add that if you have a transmitter that power several cavities and that you need to set phase for each individual cavities this is not possible in pyAML for the moment. For us, we will need this four our transition from klystron (10 cavities) to SSA (1 cavity) but this is rather for RF specialists to ensure good phase between main RF cavities and should not be a part of pyaml.
However, tuning of the phase for powered harmonic cavities may be of interest for physicists and this is already possible using pyaml. For passive harmonic cavity, we may need a dedicated cavity device if physicists need to adjust the tuning angle.
I agree with Vadim concerning this PR.

gupichon:
Got it, I'm closing it now. Thanks for the review.

PR #435, opened 2026-09-17
Allow usage of zero-length element

### LINKED ISSUE #360: Setting a value to a zero-length magnet fails silently
## The bug

Setting a value to a zero-length magnet does not raise an error. It fails silently and the readback gives nan.

As pyaml uses integrated strength, I think we should be able to deal with this case.
Or if we don't want to deal with this, at least this would be checked and forbiden at the configuration level or raise a proper error.

## How to reproduce

With a lattice with the following corrector magnet:

```
atcorrector('COR_003', 0.0, [0. 0.], 'MaxOrder', 1, 'NumIntSteps', 1, 'PolynomA', [0.], 'PolynomB', [0.]);...
```
You can write the following pyaml configuration:

```
  name: COR_003
  mapping:
    - [B0, COR_003.hcorrector]
    - [A0, COR_003.vcorrector]
  model:
    multipoles: [B0, A0]
    units: ['1', '1']
    physics:
    - ANS-C01/AE/S4-CH/currentPM
    - ANS-C01/AE/S4-CV/currentPM
 ```

You can load it into pyaml without error and select the magnet, get its strengh:
```
hcor = sr.design.magnet.get("COR_003.hcorrector")
hcor.strength.get()
>>> np.float64(-0.0)
```

But when you try to set, the value with be stuck to nan:

```
hcor.strength.set(1e-6)
hcor.strength.get()
>>> np.float64(nan)
```

gubaidulinvadim:
Closes #360 by introducing a pyAMLException when the user tries to set a zero-length element.

@GamelinAl does this answer your issue?


GamelinAl:
I am not sure. For which reasons we should throw an error instead of making it work with the integrated strength?

gubaidulinvadim:
It is so that it does not fail silently and will not confuse the user. I'd imagine you would also need the length to do the actual conversion to current. There are no thin elements in the control system.

GamelinAl:
> It is so that it does not fail silently and will not confuse the user. I'd imagine you would also need the length to do the actual conversion to current. There are no thin elements in the control system.

I think I disagree, in SOLEIL MML lattice all the corrector (skew, H/V) and even part of the sextupoles are thin magnets. I don't like it but there is no problem for unit conversion. I think we should still allow for thin magnets to be controlled to be compatible with AT lattice which uses it.

@simoneliuzzo @TeresiaOlsson What do you think?

TeresiaOlsson:
I think so too. We also had thin correctors in MML before I made them thick so could be that many labs have that and transitioning between MML and pyAML will be less easy if we don't allow it.

gubaidulinvadim:
As you wish

GamelinAl:
I proposed a solution in #360

GamelinAl:
Can we merge this @TeresiaOlsson ?

TeresiaOlsson (approved):

Issue #375, opened 2026-09-07
Feature: Add a public tuning and measurement tool holder


gupichon:
**Description, motivation and use case**
Tuning and measurement tools are currently stored in `ElementHolder`, with several default-specific properties directly on the base holder. The original holder proposal groups them under a dedicated `tool` namespace so that users can discover named tools and access defaults consistently.

This keeps the base holder focused on shared element access while providing a clear entry point for tuning workflows.

**Proposed solution**
Add a public `tool` property returning a tool holder:

- `tool.get(name)` returns the named tuning or measurement tool.
- `tool.get()` returns all configured tools as an untyped collection.
- Typed convenience properties resolve the corresponding default names: `tune`, `trm`, `orbit`, `orm`, `chromaticity`, `crm`, and `dispersion`.

Each typed property must validate its configured object type and raise a clear `PyAMLException` when its default is missing or incompatible. Keep the existing direct properties on `ElementHolder` for backward compatibility.

**Describe alternatives you've considered**
Leaving every default property on `ElementHolder` preserves existing code but makes the base API grow with each new tool. Returning raw entries from `_TOOLS` would not provide default-name mapping or type validation.

**Example**
```python
tune_correction = sr.live.tool.tune
tune_correction.set([0.31, 0.22])

tune_response = sr.live.tool.trm
tune_response.measure()

other_tool = sr.live.tool.get("OTHER_TUNE_CORRECTION")
all_tools = sr.live.tool.get()
```

**Additional context**
Parent issue: [ElementHolder API refurbishment #199](https://github.com/python-accelerator-middle-layer/pyaml/issues/199).

`ElementHolder` already has internal tool lookup and direct properties for several default tools. This issue restructures those public entry points without removing the existing API.

**Checklist**
- [x] I've assigned this issue to a project
- [ ] I've @-mentioned relevant people

Issue #373, opened 2026-09-07
Feature: Complete magnet holder navigation


gupichon:
**Description, motivation and use case**
The magnet API already has separate holders for simple magnets, combined-function magnets, and serialized magnets. The original holder proposal also calls for convenient navigation from `magnets`, including access to combined-function magnets and named magnet families.

Users should not need to know the internal holder layout when inspecting a configured magnet family interactively.

**Proposed solution**
Extend the public magnet collection API:

- Add `magnets.get_cfm()` to return all combined-function magnets as a `CombinedFunctionMagnetArray`.
- Add dynamic attribute access for configured magnet-array names when the name is a valid Python identifier, for example `magnets.QuadForTune`.
- Keep `combined_function_magnet`, `combined_function_magnets`, and the serialized-magnet holders as supported explicit APIs.

Dynamic lookup must raise `AttributeError` for unknown names and must not shadow concrete holder methods or attributes.

**Describe alternatives you've considered**
Keeping only separate CFM holders is explicit but makes the public API diverge from the navigation proposed in #199. Replacing the existing holders with one heterogeneous collection would reduce type safety and break callers that rely on typed arrays.

**Example**
```python
all_magnets = sr.live.magnets[:]
quad_family = sr.live.magnets.get("QuadForTune")
same_quad_family = sr.live.magnets.QuadForTune

combined_function_magnets = sr.live.magnets.get_cfm()
matching_quadrupoles = sr.live.magnets["Q*D*"]

one_magnet = sr.live.magnet.get("QF1E-C03")
one_magnet.strength.set(0.8)
```

**Additional context**
Parent issue: [ElementHolder API refurbishment #199](https://github.com/python-accelerator-middle-layer/pyaml/issues/199).

`MagnetsHolder` already returns `MagnetArray` objects and supports slicing and wildcard filtering. `MagnetArray` also supports its existing magnet-specific `model_name` filtering through `get_model_name()`, for example `sr.live.magnets.get()["model_name:*-H"]`. This issue adds the missing CFM entry point and ergonomic configured-family lookup without changing existing holder names.

**Checklist**
- [ ] I've assigned this issue to a project
- [ ] I've @-mentioned relevant people

Issue #372, opened 2026-09-07
Feature: Add a public diagnostic holder


gupichon:
**Description, motivation and use case**
Diagnostics are stored by `ElementHolder`, but the public API is limited to dedicated methods such as `get_betatron_tune_monitor(name)`. The original holder proposal calls for a diagnostic namespace that supports both a default monitor and named monitors.

This gives application code a stable, discoverable entry point for diagnostics without relying on private registry methods.

**Proposed solution**
Add a public `diagnostic` property returning a diagnostic holder:

- `diagnostic.get(name)` returns the named diagnostic.
- `diagnostic.get()` returns all available diagnostics as an untyped collection.
- `diagnostic.betatron_tune` resolves the `DEFAULT_BETATRON_TUNE_MONITOR` entry and validates that it is a `BetatronTuneMonitor`.

For an absent default or an incompatible configured type, raise a clear `PyAMLException`. Preserve `get_betatron_tune_monitor(name)` as a backward-compatible API.

**Describe alternatives you've considered**
Exposing `_DIAG` directly would make discovery easy but would expose mutable holder internals. Adding more direct convenience methods to `ElementHolder` would continue to grow its public surface instead of grouping related APIs.

**Example**
```python
default_tune_monitor = sr.live.diagnostic.betatron_tune
measured_tune = default_tune_monitor.tune.get()

spare_tune_monitor = sr.live.diagnostic.get("SPARE_BETATRON_TUNE_MONITOR")
all_diagnostics = sr.live.diagnostic.get()
```

**Additional context**
Parent issue: [ElementHolder API refurbishment #199](https://github.com/python-accelerator-middle-layer/pyaml/issues/199).

The base holder already stores diagnostics in its internal registry and supplies private discovery helpers for `YellowPages`. This issue adds the missing public diagnostic façade.

**Checklist**
- [ ] I've assigned this issue to a project
- [ ] I've @-mentioned relevant people

PR #449, opened 2026-09-18
Unify name/wildcard/regex lookup and miss-behavior across holders

### LINKED ISSUE #433: Feature: Unify name/wildcard access and miss-behavior across all holders
### Description, motivation and use case

Name-based lookup exists in at least six different places on the ElementHolder and array side (`ElementHolder.__getitem__`/`.find_elements()`, `GenericElementHolder.get()`, `GenericArrayHolder.get()`/`__getitem__`/`__getattr__`, `ElementArray.__getitem__`, `RFHolder.get()`, `ToolHolder.get()`), each with its own behavior: what counts as a wildcard, whether `re:` regex is supported, and what happens when nothing matches (`None`, an exception, `AttributeError`, or a silently empty or unchecked result).

This was found while adding `ToolHolder` (#375). `sr.live.bpm["BPM_001"]` raises `TypeError` since singular holders have no `__getitem__`, while `sr.live["BPM_001"]` and `sr.live.bpms["BPM_001"]` both "work", with different, disagreeing semantics from each other.

### Proposed solution

Define one shared name and wildcard resolution convention, then apply it consistently:

- Add `__getitem__` to `GenericElementHolder`, backing `.magnet`, `.bpm`, `.combined_function_magnet`, `.serialized_magnet`, with the same wildcard syntax as `ElementHolder.__getitem__`.
- Reconcile `GenericArrayHolder.__getitem__`, currently searching individual element names, with `GenericArrayHolder.get(name)`, which searches registered array names. Decide, and document, which namespace `[]` should query, since they currently disagree silently.
- Extend `RFHolder` and `ToolHolder` with the same `[]` convention.
- Uniform miss behavior, decided: a literal, exact name, no wildcard character, that does not exist must raise `PyAMLException` everywhere. That covers `ElementHolder.__getitem__`, which currently returns `None`, `GenericElementHolder`, `GenericArrayHolder`, `RFHolder` and `ToolHolder` once they gain `__getitem__`, and `find_elements()`, which currently returns an unchecked literal name with no existence check. A wildcard search that matches nothing is not an error and keeps returning an empty collection, as today.
- Wildcard trigger, decided: `*`, `?` or `[` anywhere in the key means this is an fnmatch pattern, everywhere, including `find_elements()`, which today only checks for `*` and `?` and silently treats a string containing only `[...]`, for example `"BPM_00[1-5]"`, as a literal name.
- Regex support, dropped. Team feedback (see comments) is that fnmatch already covers the vast majority of real-world naming patterns, and `re:` would only reintroduce the ambiguity and collision risk discussed above for little practical benefit. `[]` and `find_elements()` stay fnmatch only, no `re:` prefix anywhere on this side of the API. `YellowPages` keeps its own `re:` support, unaffected, tracked separately under #365.
- Case sensitivity, decided: use `fnmatch.fnmatchcase()` instead of `fnmatch.fnmatch()` everywhere a name gets matched against an fnmatch pattern. `fnmatch.fnmatch()` normalizes both the name and the pattern through `os.path.normcase()` first, which makes matching case-insensitive on Windows and case-sensitive on Linux, so the same configuration can match differently depending on the host OS. `fnmatchcase()` skips that normalization and stays case-sensitive everywhere. This applies to every existing fnmatch call site, `ElementHolder.__getitem__`, `ElementArray.__getitem__`, `find_elements()`, not only to the new ones this issue adds, `GenericElementHolder.__getitem__`, `RFHolder.__getitem__`, `ToolHolder.__getitem__`. `YellowPages` has the same case-sensitivity gap today, it is out of scope here but worth flagging for #365 too.

### Describe alternatives you've considered

Leaving each holder with its own convention avoids behavior changes but keeps the current footguns, for example `.bpms["BPMS"]` silently querying the wrong namespace, `find_elements("X[1-3]")` silently treated as literal, or `sr.live["UNKNOWN"]` silently returning `None` instead of surfacing a typo.

### Additional context

Parent issue, ElementHolder API refurbishment #199.

Found while implementing #375. This is a breaking change for any caller relying on `ElementHolder.__getitem__`'s current `None` on miss behavior, documented today at `element_holder.py:498`, `missing = sr.live["UNKNOWN"]` returning `None`. It should be called out in the changelog or release notes.

YellowPages is intentionally out of scope for this issue. #365 already covers aligning its categories with the ElementHolder sub-holders, and should be the one place deciding its own `re:` and fnmatch parity too, its scope may need to grow slightly to cover that, rather than splitting the discussion across two issues.

### Checklist

gupichon:
## Description

Name-based lookup (`__getitem__`, `.get()`, `find_elements()`) was implemented independently in at least six places across `ElementHolder` and the holder/array hierarchy, each with its own idea of what counts as a wildcard, whether `re:` regex is supported, and what happens on a miss (`None`, an exception, `AttributeError`, or a silent empty result). This surfaced while adding `ToolHolder` (#375): `sr.live.bpm["BPM_001"]` raised `TypeError` (no `__getitem__` on singular holders) while `sr.live["BPM_001"]` and `sr.live.bpms["BPM_001"]` both "worked" with disagreeing semantics. This PR defines one shared name/wildcard resolution convention (`pyaml/common/name_matching.py`) and applies it consistently to every holder and array.

## Related Issue

- Fixes #433

- [x] new feature: add `__getitem__` to `GenericElementHolder`, `RFHolder`, `RFTransmitterHolder`, `ToolHolder`, and `DiagnosticHolder` — implemented by routing all of them, plus the existing `ElementHolder.__getitem__`/`find_elements()` and `ElementArray.__getitem__`/`GenericArrayHolder.__getitem__`, through one shared resolver (`resolve_names()`/`is_wildcard()`) instead of each reimplementing its own rules.
- [x] bugfix: `fnmatch.fnmatch()` normalizes through `os.path.normcase()`, so the same wildcard pattern matched case-insensitively on Windows and case-sensitively on Linux CI — switched every call site to `fnmatch.fnmatchcase()`.
- [x] bugfix: an invalid `re:`-prefixed regex used to leak a raw `re.error`; it now raises `PyAMLException` like every other lookup failure.
- [x] decision, not code: `GenericArrayHolder.__getitem__` (e.g. `magnets["QF1*"]`) keeps searching individual element names, not registered array/family names — this is #373's recent, deliberately tested behavior. Documented explicitly instead of unifying it with `.get(name)`'s separate array/family-name namespace, since unifying would revert #373.
- [x] scope addition beyond the issue text: also added `__getitem__` to `DiagnosticHolder` and `RFTransmitterHolder`, which have the exact same shape as `RFHolder`/`ToolHolder` and didn't exist yet when the issue was written (they landed with #372 and #375).
- [x] scope addition, from thread discussion: regex (`re:` prefix) is kept, reversing what the issue currently says, and extended everywhere the resolver is used, not just `find_elements()`. A list/tuple of patterns (literal, wildcard, or regex, mixed) is also now accepted everywhere a single pattern is, unioned and de-duplicated.

- [x] `ElementHolder.__getitem__`: a literal exact name with no match now raises `PyAMLException` instead of returning `None`. Breaking change — any caller relying on the old `None`-on-miss must now catch `PyAMLException` or check membership first.
- [x] `ElementArray.__getitem__` / `GenericArrayHolder.__getitem__`: same literal-miss fix (raises instead of silently returning an empty array); wildcard trigger now also includes `[`, not just `*`/`?`; case-sensitive matching via `fnmatchcase`; gains `re:` and list-of-patterns support. A wildcard/regex pattern matching nothing is unaffected, still an empty collection, not an error.
- [x] `ElementHolder.find_elements()`: reimplemented as a thin wrapper around the shared resolver, same public behavior plus the `[`-trigger, case-sensitivity, and error-wrapping fixes above, and now also accepts a list of patterns directly.

## Testing

- [x] `tests/common/test_name_matching.py`: the shared resolver directly — literal hit/miss, wildcard hit/empty, regex hit/empty/invalid, case sensitivity, list-of-patterns (union/dedup/order, one missing literal raises)
- [x] `tests/common/test_element_holder_collection.py`: `ElementHolder.__getitem__` / `find_elements()` literal-miss-raises, `[`-as-wildcard, `re:`, list-of-patterns, case sensitivity, plus new tests combining wildcard/regex/list selections with `&`/`|`/`-` set operations (non-trivial intersection/union/difference, and the empty-result-stays-a-plain-`list` quirk under the new selection styles)
- [x] `tests/arrays/test_array_selection_types.py`: `ElementArray.__getitem__` literal-miss-raises, bracket-only wildcard, regex, list-of-patterns, case sensitivity
- [x] `tests/common/test_array_holder_navigation.py`: `GenericElementHolder.__getitem__` via `.magnet`, `.combined_function_magnet`, `.serialized_magnet`; a realistic example reproducing the `QForTest` YAML exclusion family via `[]` selection and `-`
- [x] `tests/diagnostics/test_diagnostic_accessors.py`: `DiagnosticHolder.__getitem__` and `.diagnostic.bpm[...]`
- [x] `tests/rf/test_rf_holder.py` (new): `RFHolder`/`RFTransmitterHolder.__getitem__`
- [x] `tests/tuning_tools/test_tool_holder.py` (new): `ToolHolder.__getitem__`

- [x] New and existing unit tests pass locally (`pytest tests -n auto`: 389 passed, 6 skipped; one unrelated pre-existing flake, `test_tuning_orm`, is a Windows temp-file `PermissionError` unconnected to this change)
- [x] Any existing features are not broken, aside from the explicit, documented breaking change above (literal-miss now raises instead of returning `None`)


GamelinAl:
From the test I did it seem to work.

Two small issues which seem legit when I tested:

### 1. `~re:` patterns silently return an empty array on `ElementArray` / `GenericArrayHolder`

`ElementArray.__getitem__` (`pyaml/arrays/element_array.py:609`) decides between the *name-pattern* branch and the *field-selector* branch with

```python
fields = [] if key.startswith("re:") else key.split(":")
```

Only a bare `re:` prefix is recognised. The new docstring says `~` may prefix any pattern, and `ElementHolder.__getitem__` honours that (it goes straight to `resolve_names`), but with the `~` in front the key `"~re:^QF1.*$"` is split into `["~re", "^QF1.*$"]`, `__eval_field` looks up a non-existent `get_~re` accessor, gets `""` for every element, and the result is an empty array with no error. This affects every `GenericArrayHolder.__getitem__` too (`sr.design.magnets[...]`, `sr.live.bpms[...]`, …) since they delegate here.

```python

sr = Accelerator.load("tests/config/EBSTune-patterns.yaml", ignore_external=True)
magnets = sr.design.magnets[:]           # 124 magnets

len(magnets["re:^QF1.*$"])               # 62
len(magnets["~QF1*"])                    # 62   (wildcard exclusion works)
len(magnets["~re:^QF1.*$"])              # 0    <-- expected 62, silently empty
len(sr.design.magnets["~re:^QF1.*$"])    # 0    (same path via the holder)
len(sr.design["~re:^QF1.*$"])            # 65   (ElementHolder path is correct: all elements minus QF1*)
```

Fix: ignore the `~` when deciding the branch, e.g.

```python
fields = [] if key.lstrip("~").startswith("re:") else key.split(":")
```

(or route any key starting with `~` or `re:` directly to `resolve_names`), plus a test such as `assert magnets["~re:^QF1.*$"] == magnets["~QF1*"]`.

---

### 2. `find_elements` docstring promises YAML `elements:` parity, but `_fill_array` has different `~` semantics and raises

The new `find_elements` docstring (`pyaml/common/holders/element_holder.py:323`) says `~` works "as in the `elements:` selector list of a YAML array declaration, e.g. `["QD2*", "QF1*", "~QF1E-C05"]`", and that a lone `~pattern` means "every element except those". `_fill_array` (line 376), which actually builds arrays from YAML, does not behave that way: it collects inclusions and exclusions separately and then does `all_names.remove(name)` for each excluded name, so it raises a bare `ValueError` whenever an excluded name is not among the inclusions (lone `~`, exclusion outside the included set, or two overlapping `~` entries).

```python
from pyaml.configuration.manager import ConfigurationManager

def build(elements):
    """EBS test config plus one `pyaml.arrays.magnet` declared with `elements:`."""
    manager = ConfigurationManager()
    manager.add("tests/config/EBSTune-patterns.yaml")
    manager.add({"arrays": [{"type": "pyaml.arrays.magnet", "name": "Arr", "elements": elements}]})
    return manager.build(ignore_external=True)

sr = build(["QD2*"])

for elements in (["~QF1E-C05"], ["QD2*", "~QF1E-C05"], ["QF1*", "~QF1E*", "~QF1E-C05"]):
    print(len(sr.design.find_elements(elements)))   # 126, 62, 31
    build(elements)                                 # PyAMLConfigException each time
```

| `elements` | `sr.design.find_elements(elements)` | YAML `elements:` (via `_fill_array`) |
|---|---|---|
| `["~QF1E-C05"]` | 126 names (everything but one) | `PyAMLConfigException: list.remove(x): x not in list` |
| `["QD2*", "~QF1E-C05"]` | 62 names (all QD2*) | `PyAMLConfigException: list.remove(x): x not in list` |
| `["QF1*", "~QF1E*", "~QF1E-C05"]` | 31 names | `PyAMLConfigException: list.remove(x): x not in list` |

The `_fill_array` code is pre-existing, but this PR now documents the two as equivalent, and the docstring's own example (`"~QF1E-C05"` excluding something that `QD2*`/`QF1*` happen to include) only works because the excluded name is inside the include set. Fix: make `_fill_array` use the resolver it now advertises:

```python
all_names = self.find_elements(element_names)
```

which removes the manual include/exclude bookkeeping and both `ValueError` cases, and add a test for at least the `["QD2*", "~QF1E-C05"]` shape through YAML. (Existing test `test_configured_exclusion_family_reproduced_in_a_single_getitem_call` already covers the "exclusion inside the include set" case.)

gupichon:
@GamelinAl, it should be ok now.

GamelinAl:
I found a different behavior when using `sr.design.magnet` and  `sr.design.magnets`:
image


One could also wonder why we have both now, but that's for another PR

There is also a different behavior between `sr.design.magnets` and `sr.design` which does not return the array in the same order (if the input key is not already in the "good" order):

image

gupichon:
Yes, this one will be tricky to solve. There was a brief discussion about element ordering when I implemented wildcard and regex support for arrays on the configuration side. I'm looking into it, but the agreed rule was to keep the order of the devices as they are listed in the configuration file (which should match the lattice order).

gupichon:
@GamelinAl : Ok, auto-typing is working fine. Your search query matches CombinedFunctionMagnet elements, which do not inherit from Magnet (this is confusing because they are placed alongside all the other magnets).

As for the order, that's a bug. Elements should always be returned in the configuration order.

GamelinAl:
> @GamelinAl : Ok, auto-typing is working fine. Your search query matches CombinedFunctionMagnet elements, which do not inherit from Magnet (this is confusing because they are placed alongside all the other magnets).
> 
> As for the order, that's a bug. Elements should always be returned in the configuration order.

No what I mean is that `sr.design.magnet["SH*"]` should return a `MagnetArray` not an `ElementArray`
Sorry that was not clear enough.

gupichon:
> > @GamelinAl : Ok, auto-typing is working fine. Your search query matches CombinedFunctionMagnet elements, which do not inherit from Magnet (this is confusing because they are placed alongside all the other magnets).
> > As for the order, that's a bug. Elements should always be returned in the configuration order.
> 
> No what I mean is that `sr.design.magnet["SH*"]` should return a `MagnetArray` not an `ElementArray` Sorry that was not clear enough.

You were clear. However, sr.design.magnet also contains the CFMs, which are not magnets. They are not in sr.design.magnets.

gupichon:
@GamelinAl, it's corrected:

```python
>>> sr.design.magnet["SH??-???-?"]
MagnetArray(name='', size=194, peer=Simulator:design, elements=['SH1A-C04-H', 'SH1A-C04-V', ... +191 more ..., 'SH3E-C03-S'])
>>> sr.design.magnets["SH??-???-?"]
MagnetArray(name='', size=194, peer=Simulator:design, elements=['SH1A-C04-H', 'SH1A-C04-V', ... +191 more ..., 'SH3E-C03-S'])
>>> sr.design.magnets["SH1A-C04-V", "SH1A-C04-H"]
MagnetArray(name='', size=2, peer=Simulator:design, elements=['SH1A-C04-H', 'SH1A-C04-V'])
>>> sr.design.magnets["SH1A-C04-H", "SH1A-C04-V"]
MagnetArray(name='', size=2, peer=Simulator:design, elements=['SH1A-C04-H', 'SH1A-C04-V'])
>>> sr.design.magnet["SH1A-C04-V", "SH1A-C04-H"]
MagnetArray(name='', size=2, peer=Simulator:design, elements=['SH1A-C04-H', 'SH1A-C04-V'])
>>> sr.design.magnet["SH1A-C04-H", "SH1A-C04-V"]
MagnetArray(name='', size=2, peer=Simulator:design, elements=['SH1A-C04-H', 'SH1A-C04-V'])
```

gupichon:
Sorry I mixed that up. The CFMs are in `sr.design` but not in `sr.design.magnets` or `sr.design.magnet`.

gupichon:
I found and fixed a bug along the way. This works now:

```python
>>> sr.design["SH*"] - sr.design.magnets["SH*"]
CombinedFunctionMagnetArray(name='', size=96, peer=Simulator:design, elements=['SH1A-C04', 'SH2B-C04', ... +93 more ..., 'SH3E-C03'])
```

GamelinAl:
I added a test to check for array order in the different way to access the data.
I think it's still no good for `ElementHolder.__getitem__`, `find_elements()` and the other holders ( RF, tool and diagnostic) which still keep the order of the list.
Maybe this can be fixed at the `resolve_names` level.

